PR #139, opened 2016-06-10
Initial prototype of Blocks UI

stefan-j:
Allows drawing of text, circle, rectangle
Additional blocks for translation and changing color

The server was left intact.
An additional folder funblocks-base contains the Blockly logic (with bindings) and code generation,
The web folder contains new items in js and css that were used.
web/blockworld.js is based on a subset of the origin codeworld.js. Both blocks and the regular UI can be made to use the same codebase as codeworld.js with some minor changes.
web/blockly is a git submodule that contains Blockly.
The UI for the blocks currently uses bootstrap. We should probably change this later so that both Uis have the same look and feel.

I don't have a good way of compiling funblocks-base. For now I'm just running ghcjs against is with symlinks in web/js to point to the compiled js files.

The block based UI can be accessed from /blocks.html

P.S. I'm not sure why the scrollbars are suddenly black, they should be gray.


funblocks-base/funblocks-base.cabal
```diff
@@ -0,0 +1,32 @@
...
+
+version:             0.2.0.0
+
+synopsis:            A functional block-based programming language for CodeWorld
+
+homepage:            https://github.com/stefan-j/FunBlocks
+license:             Apache-2.0
+license-file:        LICENSE
+author:              Stefan Jacholke
+maintainer:          stefanjacholke@gmail.com
```
cdsmith: Please add this line:

    Copyright:  (c) 2016, The CodeWorld Authors

web/blockly
```diff
@@ -0,0 +1 @@
+Subproject commit 04d6d119c076a42fb5777bb3f272d8256f2de499
```
cdsmith: You should add this in third_party, and symlink from here.

web/blocks.html
```diff
@@ -0,0 +1,157 @@
...
+    <script src="https://code.jquery.com/jquery-2.2.4.min.js"></script>
+
+
+    <link rel="stylesheet"
+    href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css"
+    integrity="sha384-1q8mTJOASx8j1Au+a5WDVnPi2lkFfwwEAa8hDDdjZlpLegxhjVME1fgjWPGmkzs7"
+    crossorigin="anonymous">
+
+    <!-- Optional theme -->
+    <link rel="stylesheet" href="css/bootstrap.min.css">
```
cdsmith: Why do you need both the CDN and local version of this?  I'd prefer just the CDN version.

stefan-j:
Added copyright
Moved third party files to third_party

The local bootstrap.min.css is a additional theme over bootstrap. I don't think it has a CDN and might want to be modified later.

web/js/blocks_rts.js
```diff
@@ -0,0 +1 @@
+../../funblocks-base/src/Main.jsexe/rts.js
```
cdsmith: Is this any different from the existing rts.js?  I think the RTS is always the same.

Issue #186, opened 2016-06-26
Definition blocks don't update correctly with type changes.


cdsmith:
Reproduction steps:

1. Start with https://code.world/blocks#PzkcPBeGIhH7FDLlqF8AYcg
2. Disconnect the text block from the `let` block for foo.

The `foo` block in the main expression should go back to a polymorphic block.  Instead, it just disappears!

cdsmith:
This still doesn't quite work.  The block doesn't disappear now, but it is insufficiently polymorphic.

For example, try building this program (leaving the ____ terms disconnected):

    main = drawingOf(if ____ == foo then blank else blank)
    foo = ____

Completing the definition to 'foo == 5' correctly updates types in the `==` block.  But disconnecting the foo block again leaves the type specialized in the `==` block.

cdsmith:
There are still problems here.  I think the ad hoc approach to this one is failing.  We're essentially implementing type unification here, and there are well-known approaches to doing so.  Is it time to consider this from a more formal perspective?

Examples of cases that are still broken:

1. Define "a = a" (using a let block connected to its own variable).
2. Drag out a let block for "foo".
3. Drag the "a" from the first "let" to the second, then back again.  Oops, the blocks are still assigned the same type variable!
4. Connect another "a" to some block with a monomorphic type.  Oops, "foo" is now constrained to the same type!

Another one:

1. Create a let block for "foo", and an `==` block.
2. Place the variable "foo" on one side of the equality.  It's still polymorphic.
3. Define `foo` to a number.  The first block correctly becomes monomorphic.
4. Detach the number.  Oops, the connected `foo` block remains monomorphic!

All of these failures follow the same pattern, which is that there's some missing case where you're trying to partially correct types.  I think perhaps you should first run type unification over the entire program, then rename the type variables to match the old values if you prefer.

awmorp:
The code to restore type variables when a block is disconnected currently only works for the disconnected child block and its former parent/ancestor block. It doesn't look at any other unconnected blocks. The relevant code as a starting point is here: https://github.com/stefan-j/blockly/blob/master/core/connection.js#L623

stefan-j:
@awmorp  Thanks !
I also saw you did some work with the resetting the types upon disconnect with `copyConnectionTypes_`.

cdsmith:
This seems to be obsolete.  The problem is different now, and @stefan-j is already overhauling type inference.

Issue #190, opened 2016-06-27
Add a "%" block to numbers


cdsmith:
Several functions take numbers that are best interpreted as a percent.  In these cases, it might be better for younger students if we make it clear that this is the case.

What I have in mind is to add a block that works just like the number one, except it is labeled "___ %", where the blank is validated as a number.  The codegen for the block would divide the number by 100 automatically.  We would then use this block in the ghost blocks for certain functions: `RGBA`, `gray`, and perhaps even `scaled`.

Issue #185, opened 2016-06-26
Error creating a recursive definition


cdsmith:
I attempted to create the following definition:

    foo = circle(5) & foo

1. Drag a let block for foo onto the screen.
2. Connect the & block
3. Connect the `circle` block
4. Try to connect the `foo` block.  BOOM!

    blockly_compressed.js:1231 Uncaught TypeError: Cannot read property 'id' of null
      Blockly.Events.Abstract @ blockly_compressed.js:1231
      Blockly.Events.Move @ blockly_compressed.js:1241
      Blockly.Connection.connect_ @ blockly_compressed.js:918
      Blockly.Connection.connect @ blockly_compressed.js:928
      Blockly.BlockSvg.onMouseUp_ @ blockly_compressed.js:1152
      e @ blockly_compressed.js:1414

Retrying connecting the `foo` block succeeds.  So there's something missing in the block that gets filled in on the second try.  (Of course the program hangs when it's run, but that's expected.)

Issue #170, opened 2016-06-22
FunBlocks unsaved changes warning


stefan-j:
FunBlocks should display a warning and ask confirmation when the user tries to:

- load a project when the current canvas contains unsaved changes
- start a new project when the current canvas contains unsaved changes

cdsmith:
Still needs a warning when closing the tab/window or leaving the page.

PR #195, opened 2016-06-28
Fixed let blocks, added percentage block, warn when leaving page

stefan-j:
FunBlocks now warns the user when he tries to leave the page if there are unsaved changes, as per issue #170 

Updated Blockly to handle let variable connecting and disconnecting better. 
The type of all child blocks change when the definition type changes. Child blocks also bump away if there is a type mismatch.
The type changes to polymorphic upon disconnect.
This fixes issue #186. Issue #185 is also fixed, though no additional recursion checking is done.

Added a percentage block. RGBA, Gray and Scaled also now use this percentage block as shadow blocks in the toolbox. This is work for issue #190.
The RGBA block was also fixed.

### DIFF
--- a/web/js/blockworld.js
+++ b/web/js/blockworld.js
@@ -270,8 +270,13 @@ function init()
         CodeMirror.registerHelper('hintWords', 'codeworld', hints);
     });
 
-
-
+    window.onbeforeunload = function(event) {
+        if (containsUnsavedChanges()) {
+            var msg = 'There are unsaved changes to your project. ' + 'If you continue, they will be lost!';
+            if (event) event.returnValue = msg;
+            return msg;
+        }
+    }
 
 }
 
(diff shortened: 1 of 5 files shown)


web/js/blockworld.js
```diff
@@ -351,8 +356,17 @@ function getWorkspaceXMLText()
 
 function containsUnsavedChanges()
 {
+  var blank = '<xml xmlns="http://www.w3.org/1999/xhtml"></xml>';
+  // Initialize if empty
   if(!lastXML)
-    return false;
+  {
+    lastXML = getWorkspaceXMLText();
```
cdsmith: This seems wrong.  You are changing lastXML without actually saving anything.  Don't you mean this:

    var blank = '<xml xmlns="http://www.w3.org/1999/xhtml"></xml>';
    return getWorkspaceXMLText() != (lastXML || blank);

cdsmith: I'm not 100% sure what you're doing here, so I'll merge, and then fix if it doesn't work.

stefan-j:
Thanks ! Your solution is much more elegant. I prefer it.

I can change it, though I have a rather large commit reducing code duplication and I'm unsure how to create a pull request for an entirely separate commit.

stefan-j:
BTW, I tested your version:

```
function containsUnsavedChanges()
{
  var blank = '<xml xmlns="http://www.w3.org/1999/xhtml"></xml>';
  return getWorkspaceXMLText() != (lastXML || blank);
}
```

and it works fine.

cdsmith:
I have made the change locally, and I'll commit it here.

Issue #222, opened 2016-07-18
Better local variables


cdsmith:
The existing model for local variables has some serious drawbacks.

1. Most importantly, local variables look *completely* different from built-in names.  Pedagogically, this is a disaster.  We need to find a better answer.
2. The existing system is broken anyway.  There are different blocks for references to a local variable bound by a list versus a function parameter, and they look identical but aren't interchangeable.

Ideally, the blocks that define these local variables would have some kind of a local toolbox, from which one could drag blocks for their local variables.  App inventor (which uses blockly) has something kind of like this.

![appinv](https://cloud.githubusercontent.com/assets/544744/16909597/a1aa8a26-4c88-11e6-9ff2-2944e672f99e.png)

(I even wish these labels looked like blocks instead, as in the toolbox.  But if they turned into blocks when dragged, that would be sufficient.)

stefan-j:
That is actually a good idea. The current one was meant to be temporary.

Just to reiterate, in the above app-inventor picture, the user should be able to drag x,y or touchedSprite out from the block, which should instantiate a calling block with that name?

I'll see what I can do.

stefan-j:
Ah ok, I see. It opens up a toolbar upon hover.

cdsmith:
Oh, I actually would prefer if you could just drag it out.  Too bad about the toolbar.

cdsmith:
If blockly already implements the toolbar on hover approach, maybe it would also work to have just one such bock, called "local definitions" or some such thing, and put all the blocks in there.

stefan-j:
App Inventor made their own custom Flyout, so its not officially part of Blockly. 

![localvars](https://cloud.githubusercontent.com/assets/2664883/16977257/9045db7e-4e44-11e6-86ce-be12c07b494e.png)

I'm working on something like this, for the functions as well. Calling blocks can be dragged out from the white block.

stefan-j:
One minor issue would be when the user connects them to another block, where the variable name is outside scope.. 
I should probably add a warning or disable the block if the variable name is not in scope.

cdsmith:
That looks awesome!

Yes, I guess preferably trying to connect such a block outside of its scope would just fail to connect, but an error message would be okay, too.

Issue #217, opened 2016-07-05
Program blocks can be duplicated


stefan-j:
A top level program block such as `Drawing Of` can be duplicated (only one should be allowed).
This can be done by:

- Copying the block, and then pasting the block (Ctrl-C, Ctrl-V)
- Right clicking on the block, and selecting duplicate

Issue #201, opened 2016-06-29
Renaming Let blocks is broken


stefan-j:
This issue comprises of two points:

1. Currently multiple let definition may have the same name.

2. Furthermore, the incorrect caller blocks sometimes get renamed.
For example, in the program [#P3uExjsDV-Z4LbjfLJrpptQ](https://code.world/blocks#P3uExjsDV-Z4LbjfLJrpptQ), remove the number block from the bar definition, then `foo` gets incorrectly renamed to `bar2`

stefan-j:
Also, when adding a new definition with the same name, and then renaming it, the incorrect caller blocks get renamed.

stefan-j:
Renaming of caller blocks for a let definition is temporarily disabled since PR #201

cdsmith:
This issue still seems to occur.  If I add blocks for this program:

    main = drawingOf(bar)
    bar = blank

Then drag a new "let" block from the toolbox, the existing "bar" in main is renamed to "bar2" along with the new block.  It should instead keep the same name as the old block it referred to before.

Issue #172, opened 2016-06-22
Maybe continually update generated code?


cdsmith:
I feel like it would be interesting if funblocks did continuous on-demand update of the generated code.  This would make it immediately apparent what the effect is of various block manipulations.

Questions:
1. Is this really usable on smaller screens?  Not sure.
2. What about errors?  I suppose you probably want to generate code anyway, and save the error reporting for the Run button.

PR #223, opened 2016-07-27
Updated FunBlocks - User data types

stefan-j:
Changes include:

- Addition of user data types - The user is now able to build algabraic data types. No higher kinded types currently.
- Field blocks. These can be used to instantiate local variables from functions, list comprehensions and case blocks. They fail to connect if the variable is not within scope
- GHC 8 support. Compatibility with GHC 7.10.3 is maintained.
- Blocks can be used as functions to a shallow degree. If a block contains a shadow block, it'll get removed. There is still work needed here, however I'm unsure how to proceed.
- Addition of the simulation block. Allows CodeWorld simulations to be built.
- Code is continually generated. I'm not sure what the effect is on large programs yet.

This fixes issues:

- Better local variables - Issue #222 
- Program blocks can be duplicated - Issue #217 
- Better renaming - Issue #201 
- Continuous code generation - Issue #172


codeworld-base/src/Internal/Num.hs
```diff
@@ -77,7 +77,7 @@ module Internal.Num (
 
 import qualified "base" Prelude as P
 import "base" Prelude (Bool(..), (.), (==), map)
-import Numeric
+import Numeric hiding (sqrt, pi, exp, log, logBase, sin, tan, cos, asin, atan, acos)
```
cdsmith: This looks like it will break compatibility with older `base` versions.  Is it safe, or does it need an ifdef guard?

cdsmith: Actually, I'll merge this and then change it to an explicit import of `showFFloatAlt`.  That is backward compatible, and is all that was really needed anyway.

PR #237, opened 2016-08-05
FunBlocks Update - Converge UI and add code formatting

stefan-j:
Changes:

- The blocks UI now has the same look as the text CodeWorld UI. This includes the removal of bootstrap
- Code is now formatted according to a simple style. `if`s, `case` and so on get indented. 
- Code generation is also somewhat rewritten, as a result most blocks don't need to be hard coded. Each block is converted to a AST using its .functionName value and the list of inputs.


cdsmith:
Sorry for the delay!  I was out of town for the weekend.  I'll merge and push this now.

PR #247, opened 2016-08-11
Stand alone mode

nomeata:
The stand-alone mode now works, including animation and interaction. There might be rough edges, but in principle I think it can be merged. Or at least be tested and reviewed.

I did not change the default assignment of the `ghcjs` flag, but it probably should: Whoever runs `cabal install codeworld-api` probably wants the stand-alone version, and the other version is probably always built by the `build.sh` script, which is a convenient place to specify `-fghcjs`.


nomeata:
I signed it!

Issue #305, opened 2016-09-01
Performance improvements


stefan-j:
Currently it is hard working on very large programs, as there is some slowdown on connect/disconnect. This is due to:

1. Redrawing blocks is slow.
2. Algorithm W is exponential in time. This is further worsened by Type.apply applying every substitution, which is required for fully substituting over some of the 'incomplete' types tagged.

Some ideas:
1. Only redraw blocks which had a change of type. The current behavior is to re-render the connected/disconnected blocks and their parents.
2. When composing substitutions, fully apply them. I'm not sure if this will mess up intermediate substitutions.
3. Maybe cache things that don't change too often.

(I'm opening a new issue instead of reopening #177  as the implementation details have changed)

PR #314, opened 2016-09-02
FunBlocks Update - Improve performance

### LINKED ISSUE #312: Parameter blocks keep old type even after changes
Start with this: https://code.world/blocks#PwTxyVGuDi1cNBOoh8VzEiw

Disconnect the pair from the right-hand side of the MousePress event.  The case block's connectors, and the template `state` parameter block, all correctly generalize to be polymorphic.  However, the already existing instances of the `state` parameter keep their shape as pairs.

stefan-j:
Not sure if this warrants an PR, as you still need to pull Blockly.

Changes:

- Improves performance. Only render blocks where a type change occurs, improve some other typing things as well - Issue #305
- The update renders all blocks with a type change which fixes #312


cdsmith:
PRs for new blockly versions are fine.  I'm just running `git pull && git submodule update --recursive`, so this works out.

PR #356, opened 2016-09-18
Fixes #355: Allow port to be specified via CODEWORLD_API_PORT environment variable

### LINKED ISSUE #355: I would like to run codeworld-api app with port other than 3000
Currently, applications built using `codeworld-api` run exclusively on port 3000 as indicated by the message to the terminal:

```text
Open me on http://127.0.0.1:3000/
```

I'd like to run multiple CodeWorld API apps on different ports.

rcook:
This is how I'm doing this right now. I don't know if this is an idiomatic way of overriding this value, but it works for me.

-Richard

### DIFF
--- a/codeworld-api/src/CodeWorld/Driver.hs
+++ b/codeworld-api/src/CodeWorld/Driver.hs
@@ -42,11 +42,13 @@ import           Control.Monad
 import           Control.Monad.Trans (liftIO)
 import           Data.Char (chr)
 import           Data.List (zip4)
-import           Data.Maybe (mapMaybe)
+import           Data.Maybe (fromMaybe, mapMaybe)
 import           Data.Monoid
 import qualified Data.Text as T
 import           Data.Text (Text, singleton, pack)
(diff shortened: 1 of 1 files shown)


codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -502,16 +504,23 @@ setupScreenContext (cw, ch) = do
...
+
+readPortFromEnv :: String -> Port -> IO Port
+readPortFromEnv envName defaultPort = do
+    ms <- lookupEnv envName
+    return (fromMaybe defaultPort (ms >>= readMaybe))
+
 runBlankCanvas :: (Canvas.DeviceContext -> IO ()) -> IO ()
 runBlankCanvas act = do
+    port <- readPortFromEnv "CODEWORLD_API_PORT" 3000
+    let options = (fromIntegral port) { Canvas.events =
```
cdsmith: Wow, this is hideous... not your fault, though.

PR #368, opened 2016-09-25
Improve type inference for list comprehension

### LINKED ISSUE #346: Type inference is incomplete for list comprehensions
Create a list comprehension.  Before defining the source lists, use the list comprehension variables in a monomorphic context in the first expression.

Expected: when the index variable is used as a monomorphic type, the type inference alg unifies it, so the index variable is now given the more specific type.

Actual behavior: the index variable remains fully polymorphic, despite being used in a monomorphic spot.  If, later, the use attaches a list of the wrong type to the list comprehension block, it attaches, and the program is ill-typed. 

download

stefan-j:
Doesn't generalize anymore. Fixes #346


cdsmith:
Thanks!

Issue #378, opened 2016-10-29
Client timestamps for game events


cdsmith:
Splitting from #372, idea is from @nomeata

It's unfortunate that games are basically always wrong about their local state after a UI event.  That's because the server gives the event its canonical timestamp.  But we trust the clients not to cheat, so we can have the clients set timestamps instead.

The trick is that the clients need to know a safe committed timestamp, before which no new events will be added.  That is no longer the latest server event, since there's clock skew between clients.  So clients will need to send pings with their current times, so that the server knows the true committed time, which is the oldest timestamp from any client.

nomeata:
It might be sufficient if  the client estimates the time stamp the server will give to the client’s event, (it could ask for the difference by sending a time-stamped event) and then using this difference to almost correctly assign timestamps to unconfirmed local events as they enter the queue. This way, the general server protocol does not need to change.

cdsmith:
First of all, I think it's reasonable to expect that 95% of the time, the after-the-fact change in time won't matter.  Most events will commute with events from other players, except for the occasional critical point (e.g., both players try to move to the same spot, or try to click the same object).  Most events will also commute with time steps, except in a few mechanics such as "click this before it disappears")

So, if the client estimates the server timestamp, what happens then?  Does it wait for a tenth of a second or so before applying the event locally?  I'm afraid that would be even worse!  We'd add visible latency in the 95% of the time that it doesn't matter, in exchange for avoiding a glitch in the remaining 5%.  I think it would generally be better to leave things as is.

nomeata:
There is a middle way. Assume we estinate the round trip to be one second. State is `s`. Then, after the user presses a button (`e`)and half a second passed, then we'd display state `handle e (step 0.5 s)` instead of `step 0.5 (handle e s)`. This way, whenever the events comes back and we want and have to go to `handle e (step 1 s)`, there is no jump, but rather a smooth interpolation between the anticipated local state and the  eventual committed state.

cdsmith:
This vaguely related to #385, which is a big about discrepancy of event times between client and server.  Though in that case, the client was  actually incorrect.

cdsmith:
I think I want to just push ahead with the client-side timestamp plan.  I've made a few steps in that direction recently.  Some outstanding points here:

1. Clients do not yet send pings.  The changes in commit 2dfec1099bfaa6da578fee9bf0c9fa8df22656d6 make it easy to do, though.
[...]

PR #427, opened 2017-01-20
Local timestamps

nomeata:
I gave #378 a shot. The code turned out to be not too large, which is good. A very very quick test runs, but the CPU usage was very high (possibly due to missing pings). Review and possibly merging would be nice. I will have less time over the weekend.

### DIFF
--- a/codeworld-api/src/CodeWorld/Prediction.hs
+++ b/codeworld-api/src/CodeWorld/Prediction.hs
@@ -21,10 +21,8 @@ multi-player setup.
 
 {-# LANGUAGE RecordWildCards, ViewPatterns #-}
 module CodeWorld.Prediction
-    ( Timestamp, AnimationRate, StepFun
-    , Future
-    , initFuture, currentTimePasses, currentState
-    , addEvent
+    ( Timestamp, AnimationRate, StepFun, Future
+    , initFuture, currentTimePasses, currentState, addEvent
     )
     where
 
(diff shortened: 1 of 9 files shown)


nomeata:
The prediction code as well as the network layer now allows for timestamps to be sent. We just need the driver to create a thread to send them. You have a better idea about the driver and such – would you add that?

codeworld-api/src/CodeWorld/Prediction.hs
```diff
@@ -21,109 +21,144 @@ multi-player setup.
...
-    | now - commitTime > rate
-    = serverEvent step rate ours now event (commitedRateStep step rate f)
+-- Time handling.
+--
+-- Move state forward in fixed animation rate steps, and get
+-- the timestamp as close to the given target as possible (but possibly stop short)
+timePassesBigStep :: StepFun s -> AnimationRate -> Timestamp -> TState s -> TState s
+timePassesBigStep step rate target (now, s)
+    | now + rate < target
+    = timePasses step rate target (stepBy step rate (now, s))
```
nomeata: Yes, Typo! Thanks.

nomeata:
I am right now writing a test suite for the prediction code that tries to find out that no matter in what order events come in, the resulting state is the same. That should hopefully give some assurance in this not really trivial code.

cdsmith:
Okay.  Do you want to wait on the tests to merge this pull request?  Or I'm happy to merge it once the typo is fixed, and the test can be committed separately.  Your choice.

nomeata:
Since this is the second bug I had in the code, I’ll finish the test first.

nomeata:
Ok, unsurprisingly, writing a test case uncovered some corner cases. The test suite is not yet comprehensive in any way, but I guess this can be merged.

NB: The clients still don't actually send out pings.

cdsmith:
Looks like a good start.  I'll merge this, and we can iterate from here.

PR #439, opened 2017-02-21
Add a gallery of multi-player games

nomeata:
Missing: The jumping game, the two-player Dot Grab game. The HTML could
be nicer.


codeworld-api/codeworld-api.cabal
```diff
@@ -24,12 +24,12 @@ Library
   Other-modules:       CodeWorld.Color,
                        CodeWorld.Picture,
                        CodeWorld.Event,
-                       CodeWorld.Prediction,
                        CodeWorld.Driver
   Build-depends:       base < 5,
                        cereal,
                        cereal-text,
                        codeworld-game-api,
+                       codeworld-prediction,
```
cdsmith: Do you want to call this package codeworld-prediction?  If it's intended to be useful outside of CodeWorld, it may be worth its own top-level package name.  But you're welcome to start it out here, and later once it's stabilized, you can fork to an outside package, and modify codeworld-api to use that instead.

codeworld-prediction/tests/Common.hs
```diff
@@ -0,0 +1,86 @@
+module Common where
```
cdsmith: Needs a license comment.

codeworld-prediction/src/Data/MultiMap.hs
```diff
@@ -0,0 +1,52 @@
+-- |
```
cdsmith: Needs a license comment.

web/gallery-pearl.html
```diff
@@ -0,0 +1,103 @@
+<!DOCTYPE html>
```
cdsmith: What about changing the javascript to load the JSON file obtained by replacing `.html` with `.json`.  Then we could symlink this to the base gallery, and avoid code duplication.

Your choice, though.

codeworld-api/codeworld-api.cabal
nomeata: Hmm. not sure. Hence better start like this.

web/gallery-pearl.html
nomeata: Good idea, did that. The Gallery UI with code links is not really great, though.

PR #455, opened 2017-03-20
Write angles in radians instead of degrees

shauryab98:
The definitions for `solidCircle`, `thickCircle` and `circle` call functions `sector`, `thickCircle` and `arc`. These functions take angles in radians while the circle related functions specify angle in degrees. So 360 is actually treated as 360 radians.


cdsmith:
Oops!  Thanks for the patch.

PR #463, opened 2017-04-09
Fix #457 Allow Users to Open Help in New Tab

### LINKED ISSUE #457: Pop out the guide into a new tab
The guide currently opens in a dialog covering the page.  Some people prefer to keep it open in a new browser tab.  There should be an easy way, like a pop-out button, to open the guide in a new tab instead of a dialog.

three:
I heard about your project through the Summer of Haskell site and wanted to help out. Hopefully this small change makes the help a little easier to use.

A link now appears at the top of the help allowing users to open the guide in a new page. Some additional code keeps the pages "linked" so clicking on code samples will still call `loadSample` on the other window.

Another design would have been to put a popout button on the top of the SweetAlert, but I think avoiding a top bar would help save space and keep formatting consistent between browsers. Hopefully I'm not breaking any localization rules by hardcoding the link text.

### DIFF
--- a/web/js/doc.js
+++ b/web/js/doc.js
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+window.env = parent;
 (function() {
     function linkCodeBlocks(linkable=true) {
         codeworldKeywords = {};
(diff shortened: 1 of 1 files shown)


cdsmith:
Thanks!  This is merged and live now.  I did some style tweaking.

Issue #414, opened 2016-12-19
Resizing panels separately


alphalambda:
There should be a way to resize the graphics output panel and the editor panel separately. To be able to read the code shown with a projector or smart board, students need a large font setting for the editor and a small graphics output panel. Otherwise, the graphics window leaves very little space for the editor, and students cannot see enough of the code at once.

cdsmith:
I can definitely see adding controls for changing the font size in the editor, as long as the option is shown in a subtle enough place to avoid adding too much noise for students who will rarely use it.  Willing to accept a patch for this.

cdsmith:
Fixed by editor font size, and draggable borders for the run area.

PR #464, opened 2017-04-17
Use Ctrl-Up and Ctrl-Down keys to increase/decrease font size in the …

alphalambda:
…editor panel. This is to solve the problems with projectors and smart boards, as explained in [Issue 414](https://github.com/google/codeworld/issues/414)


cdsmith:
I'd be happy to merge this if you want.  Did you close this because you don't agree to the CLA, though?  That's a pain, but it's part of the system my employer insists on as part of my being able to work on this with some company resources.

PR #465, opened 2017-04-17
Use Ctrl-Up and Ctrl-Down to change the font size in the editor panel.

alphalambda:
This solves the problem with projectors and smart boards, as described in [Issue 414](https://github.com/google/codeworld/issues/414)


cdsmith:
Thanks!

cdsmith:
This is now live.

PR #470, opened 2017-04-30
Update the dependency versions

royalharsh:
I think we should update the dependency versions in `install.sh`. Like the node version, etc.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -129,8 +129,8 @@ fi
 
 # Install GHC, since it's required for GHCJS.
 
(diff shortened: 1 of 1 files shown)


cdsmith:
Looks good if it works.  Have you checked that everything works?  That is, that GHCJS boots correctly, and CodeMirror still builds and functions right?

royalharsh:
Yeah. I did my setup of the codeworld environment using these configurations and it works.

PR #473, opened 2017-05-03
Straw man PR for improved highlighting

tgdavies:
I've fixed the off by one error and have committed using the correct email address!

We still need to decide whether triggering by cursor position is the way to go, or whether hover would be better -- and if we stick with cursor position, should being in a function name and/or next to a comma in a parameter list also highlight the function?

Plus any amount of CSS tweaking, of course.


web/js/function-highlight-addon.js
```diff
@@ -0,0 +1,176 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
```
cdsmith: If this has a third-party license, I have to ask you to move it to the third_party license, in a subdirectory with the license that applies to it.  Then symlink it from web/js

tgdavies: That file is a modified copy of a CodeMirror file, which is MIT licensed -- so my interpretation is that we can relicense it as long as we keep the copyright?

Is the third_party directory designed for files which CodeWorld is modifying, or just for straight use of dependencies?

cdsmith: Yeah, it does need to go in third_party, with the original license.  I'm not a lawyer, so I'm  just going to follow Google's rules here, rather than speculate on the ability to change licenses. :)

cdsmith: If you like, you can add a README to that directory describing your changes, and you can also add a copyright mention for "The CodeWorld Authors".  Neither is necessary.

tgdavies: Ok, Chris, will do.

cdsmith:
I think this is better than not being there.  Further improvements to styles and such can come later.

Issue #436, opened 2017-02-01
Infinite list of colors


nomeata:
Often when I program something in CodeWorld, I write something like 

    cycle [red, green, blue] !! n

or `zip` such a list with a list of objects, to color them nicely. It would be fun to have an infinite list `colors :: [Color]` provided by CodeWorld, especially if the first few colors are picked to nicely go together.

PR #475, opened 2017-05-07
Infinite list of colors

peterbecich:
Simple implementation of this request: https://github.com/google/codeworld/issues/436

A function `rainbow` takes a parameter `speed`.  With a speed of 1.0, `rainbow` produces an infinite list that cycles through the HSL color spectrum in 100 indices.  With a speed of 0.5, this cycle should occur in 200 indices (unless I'm making a trivial algebra mistake :) ).

selection_152

selection_151

selection_150

selection_149

### DIFF
--- a/codeworld-api/src/CodeWorld/Color.hs
+++ b/codeworld-api/src/CodeWorld/Color.hs
@@ -16,6 +16,8 @@
 
 module CodeWorld.Color where
 
+import Data.Functor (fmap)
+
 data Color = RGBA !Double !Double !Double !Double deriving (Show, Eq)
 type Colour = Color
 
@@ -92,6 +94,18 @@ translucent (RGBA r g b a) = RGBA r g b (a/2)
 gray, grey :: Double -> Color
 gray = grey
 grey k = RGBA k k k 1
+
+-- Produce an infinite list of colors:
+-- https://github.com/google/codeworld/issues/436
+-- The list loops through the HSL spectrum with fixed
+-- saturation and lightness:
+-- https://www.w3schools.com/colors/colors_hsl.asp
+-- A 'speed' parameter allows the user to loop through the colors
+-- in fewer indices.
+rainbow :: Double -> [Color]
+rainbow speed = let
+  coefficients = fmap (\i -> speed * i * pi / 50.0) [0..]
+  in fmap (\h -> fromHSL h 0.75 0.5) coefficients
 
 hue :: Color -> Double
 hue (RGBA r g b a)
--- a/codeworld-api/src/CodeWorld/Color.hs
+++ b/codeworld-api/src/CodeWorld/Color.hs
@@ -17,6 +17,8 @@
 module CodeWorld.Color where
 
 import Data.Functor (fmap)
+import Data.List (unfoldr)
+import Data.Fixed (mod')
 
 data Color = RGBA !Double !Double !Double !Double deriving (Show, Eq)
 type Colour = Color
@@ -95,17 +97,29 @@ gray, grey :: Double -> Color
 gray = grey
 grey k = RGBA k k k 1
 
+
 -- Produce an infinite list of colors:
 -- https://github.com/google/codeworld/issues/436
 -- The list loops through the HSL spectrum with fixed
 -- saturation and lightness:
 -- https://www.w3schools.com/colors/colors_hsl.asp
--- A 'speed' parameter allows the user to loop through the colors
--- in fewer indices.
-rainbow :: Double -> [Color]
-rainbow speed = let
-  coefficients = fmap (\i -> speed * i * pi / 50.0) [0..]
-  in fmap (\h -> fromHSL h 0.75 0.5) coefficients
+-- http://stackoverflow.com/a/7728626/1007926
+
+colorHues :: [Double]
+colorHues = let
+  start = 0.0
+  end = 2*pi
+  foldF :: (Double, Double) -> Maybe (Double, (Double, Double))
+  foldF (level, index) = let
+    pow2 = 2.0 ** (level - 1.0)
+    result = (end - start) * index / pow2
+    index' = mod' (index + pi/8) pow2
+    level' = if (index' == 1.0) then (level+1) else level
+    in Just (result, (level', index'))
+  in unfoldr foldF (1.0, 1.0)
+
+colors :: [Color]
+colors = fmap (\h -> fromHSL h 0.75 0.5) colorHues
 
 hue :: Color -> Double
 hue (RGBA r g b a)
--- a/codeworld-api/src/CodeWorld/Color.hs
+++ b/codeworld-api/src/CodeWorld/Color.hs
@@ -16,9 +16,11 @@
 
 module CodeWorld.Color where
 
-import Data.Functor (fmap)
 import Data.List (unfoldr)
 import Data.Fixed (mod')
+
+import System.Random (mkStdGen)
+import System.Random.Shuffle (shuffle')
 
 data Color = RGBA !Double !Double !Double !Double deriving (Show, Eq)
 type Colour = Color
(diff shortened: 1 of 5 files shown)


peterbecich:
I've signed the CLA

cdsmith:
Thanks for the patch.  I have a few concerns, though.

First, and most trivially, we should definitely call this something other than "rainbow".  Plenty of students will want to use that word for their own pictures.  (In fact, `grep` tells me that over a thousand programs have done that.)  Some breakage is okay between semesters, but it's better to stick with more matter-of-fact descriptive names for the standard library.  For example, `colors` would be okay.  `allColors` is also possible.

A second, and bigger, concern is that this is too complex an API.  I'm not convinced the `speed` parameter is necessary, nor that it has a strong meaning.

Finally, and most importantly, this does not seem to meet the requirements of the use cases that motivated the feature request.  The basis of this idea came from a situation where I have a game for arbitrarily many players, and want to assign each a color arbitrarily.  Then I should just be able to assign each player the color at their player number in the list, and trust that this will do something reasonable.

It's a bit more tricky to say in formal terms what that means.  Some thoughts on this:

1. A desirably property would be that the list approximately maximizes the minimum perceived difference in the first `n` elements, as a function of `n`.  That way, you could assign new players colors from the list, and would get a nice selection without two very nearby colors.  That suggests the hue should be a sequence something like: 0, 180, 90, 270, 45, 135, 225, 315, ...

2. A second desirable property, though, is that the colors are recognizable.  That can be achieved by dividing into thirds instead of halves in the first step, and halves from there on out.  So then we end up with something like: 0, 120, 240, 60, 180, 300, 30, 90, 150, 210, 270, 330, ... which hits all the primary, secondary, and tertiary colors before moving on to stranger combos.

3. A third desirable property is that adjacent distances are also maximized as much as possible.  This suggests that after the first six elements of the list above, instead of proceeding with 30, 90, 150, 210, 270, 330, the list should reorder those elements to spread out adjacent elements... something like 90, 210, 330, 150, 270, 30 (thus keeping the minimum consecutive difference to 120 degrees)

I haven't thought much about whether there's something like a unique "optimal" (in some interesting sense) sequence that best satisfies these properties.

peterbecich:
Thanks for the detailed explanation.

Is it essential that the list of colors be infinitely long?  Or can `colors` be a function that takes a size parameter?

Can we unfix the lightness and saturation parameters for some extra variety?

I will investigate this:
http://stackoverflow.com/questions/7656179/generate-visually-different-colors-with-an-unknown-color-collection-size

cdsmith:
As far as the infinite list, I'm going to say yes.  It's part of the CodeWorld philosophy that we ruthlessly simplify our APIs, even to the point of losing (non-essential) functionality.  So even if you could choose better colors (at any given `n`) with the parameter, I'd still be in favor of just one list.

Varying saturation and luminosity is fine.

peterbecich:
I've implemented this (http://stackoverflow.com/a/7728626/1007926) with one a couple of the parameters tweaked.  Saturation and luminosity remain fixed.

Seems to give a nice distribution:
```
*CodeWorld.Color Data.List> take 25 colorHues 
[6.283185307179586,2.4674011002723395,4.934802200544679,1.1190179936374323,
3.5864190939097718,6.053820194182111,2.2380359872748645,4.705437087547204,
0.889652880639957,3.3570539809122963,5.824455081184636,2.008670874277389,
4.476071974549729,0.6602877676424819,3.1276888679148214,5.59508996818716,
1.779305761279914,4.246706861552253,0.4309226546450067,2.898323754917346,
5.3657248551896854,1.549940648282439,4.017341748554778,0.20155754164753154,
2.668958641919871]

*CodeWorld.Color Data.List> sort $ take 25 colorHues 
[0.20155754164753154,0.4309226546450067,0.6602877676424819,0.889652880639957,
1.1190179936374323,1.549940648282439,1.779305761279914,2.008670874277389,
2.2380359872748645,2.4674011002723395,2.668958641919871,2.898323754917346,
3.1276888679148214,3.3570539809122963,3.5864190939097718,4.017341748554778,
4.246706861552253,4.476071974549729,4.705437087547204,4.934802200544679,
5.3657248551896854,5.59508996818716,5.824455081184636,6.053820194182111,
6.283185307179586]
```

selection_153
selection_154
selection_155
selection_156

Thanks for the feedback!

codeworld-api/src/CodeWorld/Color.hs
```diff
@@ -93,6 +97,30 @@ gray, grey :: Double -> Color
 gray = grey
 grey k = RGBA k k k 1
 
+
+-- Produce an infinite list of colors:
+-- https://github.com/google/codeworld/issues/436
```
cdsmith: I'm generally opposed to leaving breadcrumbs in comments.  Historical information like this belongs in commit logs rather than cluttering the source code.  Instead, just write a one-line description of the purpose in the exported function (`colors`)?  Better if it's a haddock comment, especially since this one is non-obvious.

codeworld-api/src/CodeWorld/Color.hs
```diff
@@ -93,6 +97,30 @@ gray, grey :: Double -> Color
...
+-- The list loops through the HSL spectrum with fixed
+-- saturation and lightness:
+-- https://www.w3schools.com/colors/colors_hsl.asp
+-- http://stackoverflow.com/a/7728626/1007926
+
+colorHues :: [Double]
+colorHues = let
```
cdsmith: This looks great, thanks!

I think we can make one of the changes I suggested pretty easily: starting with the primary and tertiary colors.  That would look something like this:

    colors :: [Color]
    colors = red : green : blue : unfoldr nextColor (1, 6)
      where nextColor :: (Int, Int) -> Maybe (Double, (Int, Int))
            nextColor (numer, denom) =
                let hue = fromIntegral numer * 2 * pi / fromIntegral denom
                    color = fromHSL hue 0.75 0.5
                    numer' = (numer + 2) `mod` denom
                    denom' = if numer' == 1 then 2 * denom else denom
                in  Just (color, (numer', denom'))

Thoughts?

codeworld-base/src/Internal/Color.hs
```diff
@@ -100,6 +101,9 @@ gray, grey :: Number -> Color
 gray = fromCWColor . CW.gray . toDouble
 grey = gray
 
+colors :: [Color]
+colors = fmap fromCWColor CW.colors
```
cdsmith: I don't mind much either way, but why `fmap` when `map` will do?

codeworld-api/src/CodeWorld/Color.hs
```diff
@@ -93,6 +97,30 @@ gray, grey :: Double -> Color
...
+  foldF :: (Double, Double) -> Maybe (Double, (Double, Double))
+  foldF (level, index) = let
+    pow2 = 2.0 ** (level - 1.0)
+    result = (end - start) * index / pow2
+    index' = mod' (index + pi/8) pow2
+    level' = if (index' == 1.0) then (level+1) else level
+    in Just (result, (level', index'))
+  in unfoldr foldF (1.0, 1.0)
+
+colors :: [Color]
```
cdsmith: I'm a bit nervous about this name.  It's a good one, but there are 950 existing programs that define `colors` at module scope, and they will all break.  This is fewer than `rainbow`, at least; and `colors` is a less annoying name to steal.  But nevertheless, I wonder if there are better names that are less likely to break anything.

Ideas (none of which I like too well):

* `allColors`
* `assortedColors`
* `variousColors`
* `miscellaneousColors`

cdsmith:
This still seems to fall short of the goal that this list fulfills all normal requirements for "a bunch of assorted colors".  For example, this doesn't look assorted at all!

https://code.world/haskell#PShk9-3L2Ydsqk-fpnPU0UQ

After the first few iterations, we get long gradients of colors in a distinct pattern.  If I randomly shuffle the numerators at each level, I get something that looks much better, having both the properties that it chooses numbers that are evenly distributed in any region of the list, and that it starts with well-spaced and recognizable colors.

https://code.world/haskell#PM8oQ9UkJhogb9vrmrE3ZAA

This is the first implementation that I'm mostly happy with.  But there are several places with consecutive colors very similar to each other.  It would be great to figure out how to derandomize the process and actually make choices that somehow maximize the uniformity of distribution in arbitrarily chosen sublists.  But I haven't figured out the math yet, so this is good enough for now!

I'm also leaning more strongly toward `assortedColors` as the name.  I feel like it expresses the goal well.

cdsmith:
@peterbecich 

Are you finishing this?  I'm happy to merge it with `assortedColors` as the name, and with the extra shuffle in the values.  Sorry for being so picky; this is one of those things that seems reasonable, but marginal enough that it's not worth doing unless the API and behavior meets a bunch of different needs.  I think we're close enough now.

peterbecich:
@cdsmith sorry for my delay.  I think you're absolutely right to be precise.  Please wait for me to take another look at this Monday night.

cdsmith:
Okay, just checking.  Thanks!  I appreciate your patience.

peterbecich:
I appreciate your patience.  Silly question; simple dependency issue.  CodeWorld builds and runs fine but I can't replicate your example here: https://code.world/haskell#PM8oQ9UkJhogb9vrmrE3ZAA

selection_165

I'm spot checking the examples from the CodeWorld guide that don't require imports, and they work.

I am attempting this running this commit: https://github.com/peterbecich/codeworld/commit/15d45ce3802932aa4310299866353a30ccb69564

What am I doing wrong?  Thanks

cdsmith:
The root path gives you an environment that hides base, and exposes an alternative base package called codeworld-base.  You want to use the /haskell path instead.

cdsmith:
Thanks!

peterbecich:
Thanks for helping me get started with the project @cdsmith !

PR #478, opened 2017-05-10
More preparations towards creation of a self-contained codeworld server image

nomeata:


gallery-code/codeworld/PEwIFcw4EXZtdP7JdytExKA.hs
```diff
@@ -0,0 +1,57 @@
+main = collaborationOf(2, initial, step, event, picture)
```
cdsmith: I can't accept a contribution to github when a student owns the copyright, without that student agreeing to the CLA and sending the pull request.  I also don't see them contributing much to CodeWorld in the long run.

Can you split these files into a local fork instead?

nomeata: Hmm, ok. I will bundle these files with the `Vagrantfile` script that builds the image then instead.

nomeata:
Updated to remove the unwanted files.

cdsmith:
Thanks

Issue #420, opened 2016-12-31
autocomplete inserts at the wrong place when used between two spaces


cdsmith:
When the characters both before and after the cursor are spaces, autocomplete inserts its word at the wrong place.  It inserts at the end of the current block of spaces, rather than at the current cursor position.

PR #479, opened 2017-05-10
Resolve issue, autocomplete inserts at the wrong place when used between two spaces #420

pranjaltale16:
This Pull Request mainly solves the issue #420. The main problem, in that case, is with the wrong value of `token`. 

Consider an example- If I put 4 spaces and then try to make insertion at position 2 using autocorrect, it add the selected value at the end of four space because of the wrong value in `token.end` with respect to the cursor position. I updated the function for that in `web/js/codeworld_shared.js`.


pranjaltale16:
I signed it!

pranjaltale16:
@cdsmith please have a look into this.

cdsmith:
Thanks!

PR #480, opened 2017-05-12
Fix image size in Guide menu, fixes overflow-x

kaustubhhiware:
The current Guide menu looks like this : 
image

Due to this, the vertical x scrollbar gets added which looks conspicuous, especially in hand held devices . The current PR makes sure the image size is that of the menu, hence improves viewability in all types of devices:
cw
image

The overflow-x scrollbar which was not needed will be removed by this PR.


kaustubhhiware:
I signed it!

cdsmith:
Thanks!

PR #488, opened 2017-05-22
Avoid rerunning build script

kaustubhhiware:
As per [the last line of `install.sh`](https://github.com/google/codeworld/blob/master/install.sh#L180), `./build.sh` is already called once . Avoid rerunning build.sh, but needs to be mentioned since someone might want to build explicitly and that detail should be mentioned.

Edits are welcome. We can also remove `./build.sh` from install.sh as an alternative.


cdsmith:
Thanks for the documentation update.  I just changed this today, because the first run of build.sh is actually doing a lot of heavy installation.

Issue #495, opened 2017-06-05
typo in acos range error message


gladstein:
> main = drawingOf (text (printed (acos (42))))

produces the error

> The **asin** function is only defined for numbers from -1 to 1.

It should say **acos**.

cdsmith:
Thanks for the report.  Pranjal's commit fixes this.

PR #496, opened 2017-06-05
Corrected error typo in base- Num.hs for acos

pranjaltale16:
With reference to the issue #495, I made a correction in the error message for acos function.


cdsmith:
Thanks!

PR #500, opened 2017-06-16
Regex shift

pranjaltale16:
This P.R. mainly refers to:
1. Shifting of regex module from Javascript to Haskell part
2. Addition of a function to save raw error message generated by ghcjs compiler

I have used `subRegex` function from `Text.Regex` library in order to replace regex patterns in Haskell. Also, the error message is stored in two files, one(hash.err.txt) here filtered message is saved and the other one(hash.rawerr.txt) the error message generated by ghcjs.


pranjaltale16:
There are some issues with this 'invalid-email-address' commits 
I am closing this P.R. and will again put a clean P.R.

PR #502, opened 2017-06-19
Regex shift1

pranjaltale16:
This P.R. mainly refers to:
Shifting of regex module from Javascript to Haskell.
I have used `subRegex` function from `Text.Regex` library in order to replace regex patterns in Haskell.


cdsmith:
Thanks for your patience!

PR #503, opened 2017-06-23
Standalone compiler for codeworld

pranjaltale16:
This P.R. Contributes to shifting of the compilation part from codeworld-server part to an independent cabal project which can be easily executed without depending on any of the other dependencies associated particularly with codeworld.

A user has to supply, source file, output file, error file, and parsing arguments.
Test case added in codeworld-compiler is added for the basic integration of test-suite.


codeworld-compiler/src/Compile.hs
```diff
@@ -0,0 +1,156 @@
...
+import           System.Directory
+import           System.FilePath
+import           System.IO
+import           System.IO.Temp (withSystemTempDirectory)
+import           System.Process
+import           Text.Regex.TDFA
+
+import ErrorSanitizer
+
+compileSource :: FilePath -> FilePath -> FilePath -> [String] -> IO Bool
```
cdsmith: Making mode a list here is definitely not necessary.  It lets details of how you parsed the command line leak into unrelated parts of the program.  Please fix.

codeworld-compiler/src/Compile.hs
```diff
@@ -0,0 +1,156 @@
...
+compileSource dir outDir errDir mode = checkDangerousSource dir >>= \case
```
cdsmith: The parameters named with "dir" don't seem to refer to directories, which is a bit confusing.  What about calling them "src", "out", and "err"?

codeworld-compiler/src/Compile.hs
```diff
@@ -0,0 +1,156 @@
...
+        return False
+    False -> withSystemTempDirectory "buildSource" $ \tmpdir -> do
+        copyFile dir (tmpdir </> "program.hs")
+        let baseArgs = case mode of
+                ["haskell"]   -> haskellCompatibleBuildArgs
+                ["codeworld"] -> standardBuildArgs
+                _           -> mode
+            ghcjsArgs = baseArgs ++ [ "program.hs" ]
+        runCompiler tmpdir userCompileMicros ghcjsArgs >>=
+          \case
```
cdsmith: Earlier, you put lambda-case on the same line.  Let's stick with that style.

codeworld-compiler/test/Main.hs
```diff
@@ -0,0 +1,7 @@
+import Data.Char
+import Control.Monad
+import System.Directory
+import Test.HUnit             -- only import needed, others are optional
+
+test1 = TestCase $ assertEqual "test upCase" "FOO" (map toUpper "foo")
```
cdsmith: Can you add at least one reasonable test case, just to get things started?

codeworld-compiler/exec/Main.hs
```diff
@@ -0,0 +1,32 @@
+{-# LANGUAGE OverloadedStrings #-}
```
cdsmith: Copyright header needed.  Copy from any other source file in the project.

codeworld-compiler/test/Main.hs
```diff
@@ -0,0 +1,7 @@
+import Data.Char
```
cdsmith: Copyright header needed.  Copy from any other source file in the project.

codeworld-server/src/Build.hs
```diff
@@ -29,118 +31,28 @@ import           System.IO.Temp (withSystemTempDirectory)
...
-    ]
-
-haskellCompatibleBuildArgs :: [String]
-haskellCompatibleBuildArgs = [
-    "-dedupe",
-    "-Wall",
-    "-O2",
-    "-package", "codeworld-api"
-    ]
+getMode :: BuildMode -> [String]
```
cdsmith: Why is this a list?

codeworld-server/src/Build.hs
```diff
@@ -29,118 +31,28 @@ import           System.IO.Temp (withSystemTempDirectory)
...
-                        _                     -> filterOutput output
-                B.writeFile (buildRootDir mode </> resultFile programId) filteredOutput
-                let target = tmpdir </> "program.jsexe" </> "all.js"
-                hasTarget <- doesFileExist target
-                when hasTarget $
-                    copyFile target (buildRootDir mode </> targetFile programId)
-                return hasTarget
-        return success
+sourceAddr :: BuildMode -> ProgramId -> FilePath
+sourceAddr mode programId = (buildRootDir mode </> sourceFile programId)
```
cdsmith: Hmm.  I personally find the original expressions clearer than these wrappers, and not much longer.

codeworld-server/src/Build.hs
```diff
@@ -29,118 +31,28 @@ import           System.IO.Temp (withSystemTempDirectory)
 import           System.Process
 import           Text.Regex.TDFA
 
-import ErrorSanitizer
 import Util
 
 compileIfNeeded :: BuildMode -> ProgramId -> IO Bool
 compileIfNeeded mode programId = do
```
cdsmith: Consider moving this function to Main, so you can delete this module?

codeworld-server/src/Main.hs
```diff
@@ -1,6 +1,8 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE PackageImports #-}
```
cdsmith: Why add PackageImports and not use it?

codeworld-compiler/exec/Main.hs
```diff
@@ -0,0 +1,32 @@
...
+{-# LANGUAGE PackageImports #-}  
+
+import System.Environment
+import System.Directory
+
+import qualified "codeworld-compiler" Compile as C
+
+main = do
+    a <- getArgs
```
iteratee: I would like to see optparse-applicative used as well.

codeworld-compiler/src/Compile.hs
pranjaltale16: Apart from the case of directly calling `compileSource` function, where we are sure about the argument is either going to be codeworld or haskell, we are not sure about the no. of arguments a user may pass in order to handle that part I used [String]. Is there any other better method to handle an unknown number of passed arguments to a function?

cdsmith: Oh, I see.  I didn't understand that "mode" might not be a mode at all, but rather just an arbitrary list of GHC arguments.  Do you need this for some reason?

pranjaltale16: In order to make it a standalone compiler it is more likely that one can also pass the list of arguments apart from just passing the args as (haskell, or codeworld). making it more open for other options too.
If that part is not of much importance now then everything can be easily shifted back from [String] thing to just String.

cdsmith: Yeah, my feeling is that passing in just the mode is fine for now.  If it seems worthwhile later on to add more complex options, the best approach would be to add a proper library for doing so.

pranjaltale16: Okay, Sure.

codeworld-compiler/codeworld-compiler.cabal
```diff
@@ -0,0 +1,62 @@
+Name:                codeworld-compiler
+Version:             0.1.0.0
+Synopsis:            ghcjs based standalone compiler for codeworld
```
cdsmith: Capitalization: it's GHCJS, and CodeWorld

codeworld-compiler/codeworld-compiler.cabal
```diff
@@ -0,0 +1,62 @@
...
+License:             Apache-2.0
+License-file:        LICENSE
+Author:              The CodeWorld Authors
+Maintainer:          pranjaltale16@gmail.com
```
cdsmith: Are you volunteering? :)  If so, standard format here would be:

    Maintainer:      Pranjal Tale <pranjaltale16@gmail.com>

codeworld-compiler/codeworld-compiler.cabal
```diff
@@ -0,0 +1,62 @@
...
+Copyright:           (c) 2017, The CodeWorld Authors
+Build-type:          Simple
+Cabal-version:       >=1.10
+
+Description:
+  The ghcjs based standalone compiler for Codeworld programming environment.
```
cdsmith: Again, capitalization: GHCJS and CodeWorld

codeworld-compiler/exec/Main.hs
```diff
@@ -0,0 +1,69 @@
...
+-}
+
+import Options.Applicative
+import Control.Monad (join)
+import Data.Monoid ((<>))
+import System.Environment
+import System.Directory
+import System.IO
+
+import "codeworld-compiler" Compile as C
```
cdsmith: I'm still confused about the use of package imports.  What other package has a Compile module?\

Also, you should always write one of:

    import qualified Compile as C

or:

    import Compile

Adding `as C` without `qualified` imports symbols BOTH as `C.foo` and `foo`, which is almost never what you wanted.

In this case, I doubt a qualified import is needed, so I'm tempted to recommend the latter.

codeworld-compiler/exec/Main.hs
```diff
@@ -0,0 +1,69 @@
...
+                                          <> help "Enter the mode of compilation" )
+        opts = info parser mempty
+
+runWithOptions :: Options -> IO ()
+runWithOptions opts = do
+    fileExists <- doesFileExist (source opts)
+    if fileExists then do
+        compileOutput <- extractSource (source opts) (output opts) (err opts) (mode opts)
+        return ()
+        else putStrLn $ "File not found:" ++ (show (source opts))
```
cdsmith: I'm surprised this works, given layout on the `do`.  Can you decrease the indent on else?  Something like this looks okay:

    if fileExists
      then do
        compileOutput <- extractSource (source opts) (output opts) (err opts) (mode opts)
        return ()
      else
        putStrLn $ "File not found:" ++ (show (source opts))

codeworld-compiler/exec/Main.hs
```diff
@@ -0,0 +1,69 @@
...
+
+extractSource :: String -> String -> String -> String -> IO Bool
+extractSource  source out err mode = do 
+    res <- C.compileSource source out err mode
+    case res of 
+        True -> return True
+        False -> do
+            errFile <- readFile err
```
cdsmith: These names are backwards!  `err` is the file, and `errFile` is the actual errors.

codeworld-compiler/exec/Main.hs
```diff
@@ -0,0 +1,69 @@
...
+                                          <> short 'm' 
+                                          <> metavar "BuildMode"  
+                                          <> help "Enter the mode of compilation" )
+        opts = info parser mempty
+
+runWithOptions :: Options -> IO ()
+runWithOptions opts = do
+    fileExists <- doesFileExist (source opts)
+    if fileExists then do
+        compileOutput <- extractSource (source opts) (output opts) (err opts) (mode opts)
```
cdsmith: Record punning is really nice for this.  If you enable the extension, you can write:

    runWithOptions Options{..} = do
        fileExists <- doesFileExist source
        if fileExists then do
            compileOutput <- extractSource source output err mode

codeworld-compiler/src/Compile.hs
```diff
@@ -30,46 +31,40 @@ import           System.Process
...
+    False -> withSystemTempDirectory "buildSource" $ \tmpdir -> do
+        copyFile src (tmpdir </> "program.hs")
         let baseArgs = case mode of
-                BuildMode "haskell"   -> haskellCompatibleBuildArgs
-                _                     -> standardBuildArgs
+                "haskell"   -> haskellCompatibleBuildArgs
+                "codeworld" -> standardBuildArgs
             ghcjsArgs = baseArgs ++ [ "program.hs" ]
-        success <- runCompiler tmpdir userCompileMicros ghcjsArgs >>= \case
+        success <-runCompiler tmpdir userCompileMicros ghcjsArgs >>= \case
```
cdsmith: Please put the space back after `<-`

codeworld-compiler/src/Compile.hs
```diff
@@ -30,46 +31,40 @@ import           System.Process
...
             Nothing -> return False
             Just output -> do
-                let filteredOutput = case mode of 
-                        BuildMode "haskell"   -> output
-                        _                     -> filterOutput output
-                B.writeFile (buildRootDir mode </> resultFile programId) filteredOutput
+                let filteredOutput = case mode of
+                        "haskell"   -> output
+                        "codeworld" -> filterOutput output
+                        _             -> output
```
cdsmith: Align the `->` please?

codeworld-server/src/Main.hs
```diff
@@ -20,6 +22,7 @@
 
 module Main where
 
+import qualified "codeworld-compiler" Compile as C
```
cdsmith: Same question: what other package defines a `Compile` and requires package imports?

codeworld-compiler/test/Main.hs
```diff
@@ -0,0 +1,31 @@
...
+import qualified "codeworld-compiler" Compile as C
+import           Data.Char
+import           Control.Monad
+import           System.Directory
+import           Test.HUnit             -- only import needed, others are optional
+
+testcaseDir :: FilePath
+testcaseDir = "codeworld-compiler/test/testcase"
+
+test1 = TestCase $ assertEqual "test upCase" "FOO" (map toUpper "foo")
```
cdsmith: I guess you've already got it on your plan to add unit tests later, so this is fine for now.

pranjaltale16: Yes, I was working on optparse-applicative thing. I will update this with test case very soon.

codeworld-compiler/test/Main.hs
```diff
@@ -0,0 +1,58 @@
...
+import           Test.HUnit             -- only import needed, others are optional
+
+testcaseDir :: FilePath
+testcaseDir = "codeworld-compiler/test/testcase"
+
+testSourceFile :: String -> FilePath
+testSourceFile testName = testcaseDir </> testName </> "source.hs"  
+
+testErrorFile :: String -> FilePath
+testErrorFile testName = testcaseDir </> testName </> "error.txt"
```
cdsmith: IMO, outputs should really go somewhere outside of the source tree.  Whatever it is should be added to .gitignore, so we don't end up checking them in.

codeworld-compiler/test/Main.hs
```diff
@@ -0,0 +1,58 @@
...
+    sourceFile <- readFile (testSourceFile testName)
+    out <- compileSource 
+        (testSourceFile testName)
+        (testOutputFile testName)
+        (testErrorFile  testName)
+        "codeworld"
+    errMsg <- readFile (testErrorFile testName)
+    return errMsg
+
+test1 = TestCase (do err1 <- (compileErrorOutput "test1")
```
cdsmith: Can you give this a more descriptive name than 'test1'?  Perhaps 'test_nakedExpression'?

codeworld-compiler/test/Main.hs
```diff
@@ -0,0 +1,58 @@
...
+                     err2 <- (compileErrorOutput "test2")
```
cdsmith: I don't know what 'test2' is.

codeworld-compiler/test/Main.hs
```diff
@@ -0,0 +1,58 @@
...
+        (testSourceFile testName)
+        (testOutputFile testName)
+        (testErrorFile  testName)
+        "codeworld"
+    errMsg <- readFile (testErrorFile testName)
+    return errMsg
+
+test1 = TestCase (do err1 <- (compileErrorOutput "test1")
+                     err2 <- (compileErrorOutput "test2") 
+                     assertEqual "Test case 1" err1 err2)
```
cdsmith: Didn't you mean to read the file `savedErrorOutput "test1"`, and compare with that instead?

codeworld-compiler/test/Main.hs
cdsmith: It would be nice if you could just iterate over all subdirectories of the test case dir, and automatically run the test for each one.

codeworld-compiler/test/testcase/test1/saved_error.hs
```diff
@@ -0,0 +1,2 @@
+program.hs:1:1: error:
```
cdsmith: I think all of these file names are wrong...

pranjaltale16: Yes, actually I was completing the part and all of a sudden my system started creating some problem, and I just pushed the code in between.
I'll just update it.

pranjaltale16: Sorry for this.

codeworld-compiler/test/Main.hs
pranjaltale16: Yes, for this I am going to add a list of subdirectories and then going to iterate over that.
```
> test2 = TestList [ "test upCase"   ~: "FOO" ~=? (map toUpper "foo")
>                          , "test downCase" ~: "baR" ~=? (map toLower "BAR")
>                          ]
```
Similar to this.

cdsmith: You could do that, but then there's the possibility of adding a test to the directory, and forgetting to add the line of code to enable it.  Why not just fetch the list of subdirectories in the test code?

codeworld-compiler/test/Main.hs
```diff
@@ -53,6 +56,24 @@ compileErrorOutput testName = do
...
+                      ]
+genTestCases (x:xs) = [ x ~: do  err1 <- (compileErrorOutput x)
+                                 err2 <- (savedErrorOutput   x) 
+                                 assertEqual x err1 err2
+                      ] ++ genTestCases xs
+
+getTestCases :: FilePath -> IO Counts
+getTestCases path = do
+    dirContent <- getDirectoryContents path
+    let filtered = [ x | x <- dirContent, x `notElem` ["..","."]]
```
cdsmith: let filtered = filter (`notElem` ["..", "."]) dirContents

seems cleaner.

codeworld-compiler/test/Main.hs
```diff
@@ -53,6 +56,24 @@ compileErrorOutput testName = do
     errMsg <- readFile (testErrorFile testName)
     return errMsg
 
-test1 = TestCase (do err1 <- (compileErrorOutput "test1")
-                     err2 <- (compileErrorOutput "test2") 
-                     assertEqual "Test case 1" err1 err2)
+genTestCases :: [t] -> [Test]
```
cdsmith: This function is just a map, with an error for the empty case.  I'd want to see this function:

    toTestCase x = x ~: do
        err1 <- compileErrorOutput x
        err2 <- savesErrorOutput x
        assertEqual x err1 err2

Then you can just map that over the list, and then add a new test to fail if the list was empty?

codeworld-compiler/test/Main.hs
```diff
@@ -0,0 +1,79 @@
...
+        (testSourceFile testName)
+        (testOutputFile testName)
+        (testErrorFile  testName)
+        "codeworld"
+    errMsg <- readFile (testErrorFile testName)
+    return errMsg
+
+genTestCases :: [t] -> [Test]
+genTestCases []  = error "Empty directory please add testcase"
+genTestCases [x] =    [ x ~: do  err1 <- (compileErrorOutput x)
```
cdsmith: This function is just a map, with an error for the empty case.  I'd want to see this function:

    toTestCase x = x ~: do
        err1 <- compileErrorOutput x
        err2 <- savesErrorOutput x
        assertEqual x err1 err2

Then you can just map that over the list, and then add a new test to fail if the list was empty?

codeworld-compiler/test/Main.hs
```diff
@@ -0,0 +1,77 @@
...
+
+toTestCase x = x ~: do
+    err1 <- compileErrorOutput x
+    err2 <- savesErrorOutput x
+    assertEqual x err1 err2
+
+getTestCases :: FilePath -> IO Counts
+getTestCases path = do
+    dirContent <- getDirectoryContents path
+    let filtered = filter (`notElem` ["..", "."]) dirContents
```
cdsmith: typo here

cdsmith:
Thanks, Pranjal.  I'm merging this now.

PR #504, opened 2017-06-24
Introduces a shareable directory structure

parvmor:


cdsmith:
This seems to prevent users from accessing any existing projects.  The server gives an error about the '.cw' file being the wrong type.  Have you thought about the migration path?

parvmor:
For now I have added the code to migrate the users existing code in their root dir. Is it okay or do you want a way so that they can reorganize their stuff?

cdsmith:
This looks fine.  So long as people don't lose their existing projects!  If you think this is an improvement as is, I'll look it over and consider merging tonight.

cdsmith:
Hmm, I'm having trouble testing these except on the live site.  I'll have to figure that out before too much longer.

Styles are a bit weird, but I'm working on fixing it.

cdsmith:
I pushed commit 6d93aee7192a215c5dfc2f07177d899aaee54870 to fix styles.

PR #509, opened 2017-07-04
Allow users to inspect pictures in Debug Mode

three:
This PR adds a way of inspecting pictures in Debug Mode which is activated by clicking the Inspect button when the program is being run. Clicking on parts of a picture in debug mode will show the "stack" of what the piece is and which ways it was transformed. This will also cause parts of the user's code to be highlighted indicating where the creation and transformations took place.

Finding source locations is accomplished via GHC stack traces which are now stored in each Picture. All functions for creating and transforming pictures is given `HasCallStack` so the caller's location is stored in the CallStack. Finding the top piece of picture containing a point is done by calling `ctx.isPointInPath` for each component piece until one is hit. This information is provided by a callback registered by the entrypoint to the javascript handling the UI.

Currently, debug mode only works for the drawingOf entrypoint, but I plan to add support for animationOf, simulationOf, interactionOf and possibly collaborationOf as well.


codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -50,7 +50,7 @@ import           Control.Exception
 import           Control.Monad
 import           Control.Monad.Trans (liftIO)
 import           Data.Char (chr)
-import           Data.List (zip4)
+import           Data.List (zip4,find)
```
cdsmith: [nit] add a space after the comma, please?

cdsmith:
Thanks!  Merged.

PR #510, opened 2017-07-04
UpdateTestcase

pranjaltale16:
Added some more base test cases. 
Updated `test.sh` such that, one can just append the name of the folder in the `dirList` and it will run tests for all the registered directories.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -16,7 +16,9 @@
 
 source base.sh
 
-#Test for codeworld-compiler
-run codeworld-compiler cabal configure --enable-tests
-run codeworld-compiler cabal build 
-run codeworld-compiler cabal test
+for i in $(grep -l -i Test-Suite */*.cabal); do
+    TARGET=$(dirname $i)
+    run $TARGET cabal configure --enable-tests
+    run $TARGET cabal build
+    run $TARGET cabal test
+done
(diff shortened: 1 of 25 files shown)


test.sh
```diff
@@ -16,7 +16,10 @@
 
 source base.sh
 
-#Test for codeworld-compiler
-run codeworld-compiler cabal configure --enable-tests
-run codeworld-compiler cabal build 
-run codeworld-compiler cabal test
+DirList="codeworld-compiler"
```
cdsmith: If you wanted to be even fancier, you could try:

    for i in $(grep -l -i Test-Suite */*.cabal); do
        TARGET=$(basename $i)
        run $TARGET cabal configure --enable-tests
        run $TARGET cabal build
        run $TARGET cabal test
    done

What do you think?

pranjaltale16: The problem with base case is, it is resulting in `directory-name.cabal` while run function needs the name of the directory, so I updated it with `cut`. Is it okay?

cdsmith: Oops, my mistake!  I meant `dirname`.  Your version works, but `dirname` is probably clearer.

pranjaltale16: Done, please have a look.

cdsmith:
Thanks!

PR #512, opened 2017-07-04
Video exports with WebM streams

venkat24:
* Video recording with the HTML5 MediaRecorder API
* Start and Stop Recording buttons to capture output and download
* Runs on Chrome and Firefox

### DIFF
--- a/web/js/codeworld.js
+++ b/web/js/codeworld.js
@@ -158,9 +158,55 @@ function initCodeworld() {
     }
 }
 
-var cStream;
-var recorder;
-var chunks = [];
+class CanvasRecorder {
+    constructor(canvas, framerate) {
+        var cStream = canvas.captureStream(framerate);
+
+        this.chunks = [];
+        this.recorder = new MediaRecorder(cStream);
+        this.recorder.ondataavailable = this.addChunk(this.chunks);
+        this.recorder.onstop = this.exportStream(this.chunks);
+    }
+
+    addChunk(chunks) {
+        return function(e) {
+            chunks.push(e.data);
+        }
+    }
+
+    exportStream(chunks) {
+        return function() {
+            var blob = new Blob(this.chunks);
+
+            // Reset data
+            chunks = [];
+
+            // Set file name
+            var d = new Date();
+            var videoFileName = 'codeworld_recording_'
+                                + d.toDateString().split(' ').join('_') + '_'
+                                + d.getHours() +':'+ d.getMinutes() +':'+ d.getSeconds()
+                                +'.webm';
+
+            // Create a new video link
+            var a = document.createElement("a");
+            document.body.appendChild(a);
+            a.style = "display: none";
+
+            // Save the video
+            var url = window.URL.createObjectURL(blob);
+            a.href = url;
+            a.download = videoFileName;
+            a.click();
+            window.URL.revokeObjectURL(url);
+
+            // Remove the video link
+            a.remove();
+        }
+    }
+}
+
+var canvasRecorder;
 
 function captureStart() {
     var iframe = document.querySelector('#runner');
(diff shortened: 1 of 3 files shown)


web/js/codeworld.js
```diff
@@ -158,6 +159,63 @@ function initCodeworld() {
...
+var cStream;
+var recorder;
+var chunks = [];
+
+function captureStart() {
+    var iframe = document.querySelector('#runner');
+    var innerDoc = iframe.contentDocument || iframe.contentWindow.document;
+
+    var canvas = innerDoc.querySelector('#screen');
+    var context = canvas.getContext('2d');
```
cdsmith: Why get this context if it's not used?

web/js/codeworld.js
```diff
@@ -96,6 +96,7 @@ function init() {
         setCode('');
         if (!signedIn()) help();
     }
+
```
cdsmith: accidental blank line?

web/js/codeworld.js
```diff
@@ -158,6 +159,63 @@ function initCodeworld() {
...
+    var a = document.createElement("a");
+    document.body.appendChild(a);
+    a.style = "display: none";
+
+    url = window.URL.createObjectURL(blob);
+    a.href = url;
+    a.download = filename;
+    a.click();
+    window.URL.revokeObjectURL(url);
+}
```
cdsmith: Should you remove the anchor from the document?

web/js/codeworld.js
```diff
@@ -158,6 +159,63 @@ function initCodeworld() {
...
+                        +'.webm';
+    downloadFile(blob, videoFileName);
+}
+
+function downloadFile(blob, filename) {
+    var a = document.createElement("a");
+    document.body.appendChild(a);
+    a.style = "display: none";
+
+    url = window.URL.createObjectURL(blob);
```
cdsmith: `var url`?

web/js/codeworld.js
```diff
@@ -585,9 +643,13 @@ function run(hash, dhash, msg, error) {
     if (hash) {
         window.location.hash = '#' + hash;
         document.getElementById('shareButton').style.display = '';
+        if (!!navigator.mediaDevices.getUserMedia) {
```
cdsmith: At this point, the program may still have failed to build.  Perhaps you want to add this logic at the same place that `runner.contentWindow` is shown or hidden below.

cdsmith: Don't you want to remove the changes in this block?  They'll be undone by the code below, anyway.

web/js/codeworld.js
```diff
@@ -585,19 +642,27 @@ function run(hash, dhash, msg, error) {
...
+            document.getElementById('startRecButton').style.display = '';
+        }
     } else {
         window.location.hash = '';
         document.getElementById('shareButton').style.display = 'none';
+        document.getElementById('startRecButton').style.display = 'none';
```
cdsmith: Same here; this isn't needed.

web/js/codeworld.js
```diff
@@ -158,6 +158,63 @@ function initCodeworld() {
     }
 }
 
+var cStream;
```
cdsmith: I think it would be nicer if you encapsulated all of these global variables into a class.

web/js/codeworld.js
```diff
@@ -158,6 +158,63 @@ function initCodeworld() {
...
+var recorder;
+var chunks = [];
```
cdsmith: Doesn't this need to be cleared every time you start a recording?  It looks like the way this is now, you'll keep building more onto the data stream, but re-save the previous recorded content.

web/js/codeworld.js
```diff
@@ -158,6 +158,63 @@ function initCodeworld() {
...
+
+function exportStream(e) {
+    var blob = new Blob(chunks);
+    chunks = [];
+
+    var d = new Date();
+    var videoFileName = 'codeworld_recording_'
+                        + d.toDateString().split(' ').join('_') + '_'
+                        + d.getHours() +':'+ d.getMinutes() +':'+ d.getSeconds()
+                        +'.webm';
```
cdsmith: You haven't specified the format in options to MediaRecorder.  Are you sure the stream is in webm format?  I don't know this API too well.

web/js/codeworld.js
```diff
@@ -585,19 +642,27 @@ function run(hash, dhash, msg, error) {
     if (hash) {
```
cdsmith: I believe that somewhere in `run`, you want to stop the current recording, if there is one?

web/js/codeworld.js
venkat24: I reset chunks to to an empty array after the recording is saved, but I'll clean this up and put it in a class instead of a global.

web/js/codeworld.js
venkat24: The blob takes `video/webm` MIME type automatically since that's the default HTML video type. I've tested it.

web/js/codeworld.js
venkat24: Added this check in `de579e4`

cdsmith:
Merged.  Thanks!

Issue #521, opened 2017-07-10
Inspection mode doesn't work correctly for thick/outline shapes


cdsmith:
For instance, try the inspect button here: https://code.world/#P4AE8FvHY8-Ytvs6jK32x4A

No matter where you click, you get a popup suggesting the rectangle.  The rectangle should be selected when you click on the visible part of the rectangle, not when you click anywhere inside!

@three

Issue #517, opened 2017-07-06
Basic styling for inspect mode popup


cdsmith:
The inspect mode popup doesn't look great.  A bit of CSS would go a long way here.

cdsmith:
@three

Issue #516, opened 2017-07-06
Inspect-mode popup appears off the screen


cdsmith:
The new inspect feature works great for most shapes!  However, for shapes near the edge of the browser window, the popup window extends off the side of the window.  The position should be bumped so that it doesn't extend off the screen.

cdsmith:
@three

PR #529, opened 2017-07-28
Add debugmode support to animationOf, simulationOf and interactionOf

three:
This PR adds support for debugmode to animationOf, simulationOf and interactionOf. The animationOf and simulationOf entry must be paused before inspecting. The interactionOf entrypoint will automatically pause by clicking Inspect and will not receive events while paused. This PR should also fix bugs #521, #516 and #517.

The `isPointInStroke` function is used for strokes, consistent with `isPointInPath` for solid shapes. Line numbers in infobox are better aligned and names link to their documentation. The inspect button will only show if available. Pausing was originally implemented by holding an MVar that was released when leaving debugmode, but is now implemented as wrappers around states and events that is cleaner and requires less changes. An unreported bug where text couldn't be properly inspected was fixed.

### DIFF
--- a/codeworld-api/src/CodeWorld/Driver.hs
+++ b/codeworld-api/src/CodeWorld/Driver.hs
@@ -96,11 +96,9 @@ import           JavaScript.Object
 import           JavaScript.Web.AnimationFrame
 import qualified JavaScript.Web.Canvas as Canvas
 import qualified JavaScript.Web.Canvas.Internal as Canvas
-import qualified JavaScript.Web.Canvas.ImageData as ImageData
 import qualified JavaScript.Web.Location as Loc
 import qualified JavaScript.Web.MessageEvent as WS
 import qualified JavaScript.Web.WebSocket as WS
-import qualified JavaScript.TypedArray as TypedArray
 import           System.IO.Unsafe
 import           Unsafe.Coerce
 
--- a/codeworld-api/src/CodeWorld/Driver.hs
+++ b/codeworld-api/src/CodeWorld/Driver.hs
@@ -370,7 +370,7 @@ findTopPicture ctx ds pic = case pic of
         Canvas.font (fontString sty fnt) ctx
         width <- Canvas.measureText (textToJSString txt) ctx
         let height = 25 -- height is constant, defined in fontString
-        withDS ctx ds $ Canvas.rect ((-0.5)*width) (0.5*height) width height ctx
+        withDS ctx ds $ Canvas.rect ((-0.5)*width) ((-0.5)*height) width height ctx
         contained <- js_isPointInPath 0 0 ctx
         if contained
             then return (Just [pic])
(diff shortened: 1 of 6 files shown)


web/js/debugmode.js
```diff
@@ -111,6 +112,38 @@ function initDebugMode(getStackAtPoint) {
     });
 }
 
+function createSrcLink(pic) {
+    var li = document.createElement("li");
+
+    var preludeLink = "/doc/Prelude.html#v:" + pic.name;
```
cdsmith: Interesting idea, but I actually want to get rid of links to Haddock, because they contain a lot of Haskell-isms that CodeWorld should ignore.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -378,13 +404,39 @@ findTopPicture ctx ds pic = case pic of
...
+foreign import javascript unsafe "window.debugMode"
+    js_isDebugModeActive :: IO Bool
+
+foreign import javascript unsafe "startDebugMode()"
+    js_startDebugMode :: IO ()
+
+foreign import javascript unsafe "stopDebugMode()"
+    js_stopDebugMode :: IO ()
+
+foreign import javascript unsafe "parent.sweetAlert($2,$3,$1);"
```
cdsmith: I'd like to move the API boundary a bit further back.  The fact that we're using SweetAlert definitely is a detail of the environment, and shouldn't be fixed that way in external code.  So please define a function in the parent frame, and call that, rather than generating JavaScript that assumes SweetAlert is loaded.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -1161,7 +1214,64 @@ run initial stepHandler eventHandler drawHandler = do
...
+        sendEvent event
+
+    let picIfUnpaused state = case isPaused state of
+            True  -> drawHandler state
+            False -> pictures []
+        checkPaused = do
+            paused <- isPaused <$> getState
+            when (not paused) $ do
+                setDebugModeActive False
+                sweetAlert "error" "Sorry!" "Please pause before inspecting."
```
cdsmith: If you auto-pause interactions, why not do the same with simulations and animations?  That would be much more convenient for the user.

three:
Since sweetalert was only used for giving an error when inspecting something unpaused, it's no longer necessary now that it autopauses, so I didn't need to make an API further back. I'll keep that in mind for the next time though!

Interactions have to be handled separately than animations and simulations since there's not a built-in pause button. As of what I posted simulations and animations will now autopause on inspect and turn off inspect on play, but pausing won't automatically turn on inspect. I don't think Inspect should always be on when an animation is paused, but it might also be frustrating seeing a paused animation and being unable to inspect it. I thought the alert helped a little to explain this, but it's certainly a bit inconvenient.

I also set up http://codeworld.threedot14.com/ which has the last commit I posted running. It's a bit slow, but it should show what I'm talking about.

cdsmith:
Thanks!  Merged.

PR #536, opened 2017-08-25
Separating errorSanitizer logic to parse runtime errors also

### LINKED ISSUE #505: Error message rewriting is lost for runtime errors.
We've lost error message rewriting in runtime errors, as part of migrating the rewriting code to the server.  For example, when I lose at https://code.world/#PWCzFEaAgOuBeBUKcEPiVzw, I get:

    GAME OVER
      error, called at Line 34, Column 12 in main:Main

There used to be rules that would replace "main:Main" with "your program", and also "CallStack (from HasCallStack):" with "When evaluating:".  Now that those are bundled into the compiler, they do not get applied to runtime errors. Oops!

The solution is going to be a little ugly.  Perhaps it's to move error sanitizing into a runtime library, which can be compiled by both GHCJS and GHC?

pranjaltale16:
fix #505


pranjaltale16:
I'm getting some errors with sanitizing. I added functions for that, build is running perfectly but messages are not getting sanitized. 
@cdsmith could you please help me with this, where am I going wrong?

build.sh
```diff
@@ -23,9 +23,10 @@ run .  cabal update
 # Install the codeworld-base and codeworld-api packages
 
 run .  cabal_install --ghcjs ./codeworld-prediction \
+                             ./codeworld-error-sanitizer \
                              ./codeworld-api \
                              ./codeworld-base \
-                             ./codeworld-game-api
+                             ./codeworld-game-api
```
cdsmith: [nit] No need for the trailing space.

codeworld-api/codeworld-api.cabal
```diff
@@ -27,6 +27,8 @@ Library
                        CodeWorld.Driver
                        CodeWorld.CollaborationUI
   Build-depends:       base                 >= 4.8 && < 5,
+                       bytestring,
```
cdsmith: Can you add version bounds for bytestring, please?

codeworld-api/codeworld-api.cabal
```diff
@@ -27,6 +27,8 @@ Library
...
+                       codeworld-error-sanitizer,
```
cdsmith: This should be either here or in the impl(ghcjs) section, not both.  (Although, see the other comment about rearranging dependencies)

codeworld-api/src/CodeWorld.hs
```diff
@@ -121,7 +121,8 @@ module CodeWorld (
     pattern PointerMovement,
 
     -- * Debugging
-    trace
+    trace,
+    addMessage
```
cdsmith: Can you add a CodeWorld.Internal package, and only export from that?  I don't really want this to appear as part of the public CodeWorld API.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -1568,6 +1571,9 @@ trace msg x = unsafePerformIO $ do
 reportError :: SomeException -> IO ()
 reportError e = js_reportRuntimeError True (textToJSString (pack (show e)))
 
+addMessage :: SomeException -> IO ()
+addMessage e = js_reportRuntimeError True (textToJSString (pack (C.unpack (filterOutput (C.pack (show e))))))
```
cdsmith: Actually, this isn't so much what I had in mind.  I was thinking of something more like:

In `codeworld-api:CodeWorld.Internal`:

    addMessage :: Text -> IO ()

    run :: (Text -> IO ()) -> normal parameters -> IO ()
    run messageHandler params = stuff `catch` handle
      where handle exc = do
                messageHandler (pack (show e))
                setErrorState

In the `codeworld-api:CodeWorld`, things like:

    interactionOf :: normal parameters -> IO ()
    interactionOf params = run addMessage params

And in `codeworld-base`:

    interactionOf params = CodeWorld.Internal.run messageHandler params
      where messageHandler = addMessage . sanitizeMessage

Then ErrorSanitizer would be a dependency of `codeworld-base`, not `codeworld-api`.

PR #539, opened 2017-08-30
Fix issue #505

### LINKED ISSUE #505: Error message rewriting is lost for runtime errors.

pranjaltale16:
This PR fix #505.


pranjaltale16:
I think still exceptions are caught in codeworld-api only.
Changes are made but I'm not getting where is this going wrong. Messages are not getting sanitized.

codeworld-error-sanitizer/Setup.hs
```diff
@@ -0,0 +1,2 @@
+import Distribution.Simple
```
cdsmith: Please add a license comment here.  Copy it from any other Haskell source file.

codeworld-error-sanitizer/codeworld-error-sanitizer.cabal
```diff
@@ -0,0 +1,30 @@
...
+Author:              Chris Smith
+Maintainer:          cdsmith@gmail.com
+Copyright:           (c) 2017, The CodeWorld Authors
+Build-type:          Simple
+Cabal-version:       >=1.10
+
+Description:
+  Regex based error filter library for CodeWorld compiler 
+
+library
```
cdsmith: [nit] Please capitalize "library"

codeworld-base/src/Internal/CodeWorld.hs
```diff
@@ -57,7 +68,7 @@ simulationOf (initial, step, draw) = do
     rs <- chooseRandoms
     CW.simulationOf (initial rs)
                     (\dt w -> step (w, fromDouble dt))
-                    (toCWPic . draw)
+                    (toCWPic . draw) `catch` reportError
```
cdsmith: Please put the catch on the next line (but indented relative to the `do` block).

codeworld-base/src/Internal/CodeWorld.hs
```diff
@@ -69,7 +80,7 @@ interactionOf (initial, step, event, draw) = do
     CW.interactionOf (initial rs)
                      (\dt w -> step (w, fromDouble dt))
                      (\ev w -> event (w, fromCWEvent ev))
-                     (toCWPic . draw)
+                     (toCWPic . draw) `catch` reportError
```
cdsmith: Catch on the next line (same as above)

codeworld-base/src/Internal/CodeWorld.hs
```diff
@@ -85,11 +96,14 @@ collaborationOf (players, initial, step, event, picture) =
         (initial . randomsFrom)
         (\dt state -> step (state, fromDouble dt))
         (\player ev state -> event (state, fromCWEvent ev, fromInt player + 1))
-        (\player state -> toCWPic (picture (state, fromInt player + 1)))
+        (\player state -> toCWPic (picture (state, fromInt player + 1))) `catch` reportError
```
cdsmith: Catch on the next line (same as above)

codeworld-error-sanitizer/Setup.hs
```diff
@@ -1,2 +1,21 @@
+{-# LANGUAGE LambdaCase        #-}
```
cdsmith: You don't need these language pragmas.

codeworld-base/src/Internal/CodeWorld.hs
```diff
@@ -68,7 +68,8 @@ simulationOf (initial, step, draw) = do
     rs <- chooseRandoms
     CW.simulationOf (initial rs)
                     (\dt w -> step (w, fromDouble dt))
-                    (toCWPic . draw) `catch` reportError
+                    (toCWPic . draw)
+                    `catch` reportError
```
cdsmith: Sorry, I should have been clearer.  I'm trying to avoid the catch looking like a parameter, so let's please indent less than the parameters.  Four spaces past the do block indent would be consistent.

codeworld-error-sanitizer/Setup.hs
pranjaltale16: Sorry for this.

cdsmith:
Thanks!

Issue #541, opened 2017-08-31
codeworld-compiler tests fail


cdsmith:
@pranjaltale16 

Unfortunately, these only run when someone runs the script by hand.  I fixed the test runner script, but the tests are failing.  From a quick glance, some of the problems seem to be due to unicode characters, but there are other failures, too.

pranjaltale16:
The main issue is with unicode only.
For ex, In first case:
```
### Failure in: 0:test_variableName        
test/Main.hs:69
test_variableName
expected: "\nprogram.hs:1:18: error:\n    \8226 Variable not in scope: circl :: Picture\n    \8226 Perhaps you meant one of these:\n        circ (line 2), circle (defined in the standard library)\n"
 but got: "\nprogram.hs:1:18: error:\n    \\8226 Variable not in scope: circl :: Picture\n    \\8226 Perhaps you meant one of these:\n        circ (line 2), circle (defined in the standard library)\n"
```

Also, some of the messages will also change once we integrate haskell-src-exts, I think, it is better to update messages once we integrate it.

PR #542, opened 2017-08-31
haskell-src-exts support in codeworld-compiler

pranjaltale16:
This PR adds haskell-src-exts support in codeworld-compiler to parse source code and look for errors from the generated tree.


pranjaltale16:
Please have a look and suggest changes and improvements that can be done.
Also, can we keep the commented code, or should I remove that?

codeworld-compiler/src/ParseCode.hs
```diff
@@ -0,0 +1,128 @@
...
+-}
+
+{-
+badExps (App x (Lit m _) (Lit m _))  = []
+badExps (App x (Lit m _) (Var m _))  = []
+-}
+badExps :: Exp SrcSpanInfo -> [String]
+badExps (App x (Var l _) e) | not (isGoodExpRhs e) = [
+        "Missing parantheses in function applicaion at: program.hs:"
+        ++ (show $ parseSrcSpanEndLine l) ++ ":"
```
cdsmith: I think you should write this function:

    formatLocation :: SrcSpanInfo -> String
    formatLocation (SrcSpanInfo s _) = name ++ ":" ++ show line ++ ":" ++ show col
        where name = srcSpanFilename s
              line = srcSpanStartLine s
              col = srcSpanStartColumn s

Then you can use that for any line and column references.

codeworld-compiler/src/ParseCode.hs
```diff
@@ -0,0 +1,128 @@
...
+badInfixExps _                                              = []
+-}
+
+{-
+badExps (App x (Lit m _) (Lit m _))  = []
+badExps (App x (Lit m _) (Var m _))  = []
+-}
+badExps :: Exp SrcSpanInfo -> [String]
+badExps (App x (Var l _) e) | not (isGoodExpRhs e) = [
+        "Missing parantheses in function applicaion at: program.hs:"
```
cdsmith: To match existing errors, let's format this as:

    program.hs:12:5: error: Missing parentheses in function application

(Also, note the correct spelling of "parentheses")

codeworld-compiler/src/ParseCode.hs
```diff
@@ -0,0 +1,128 @@
...
+        ++ (show $ parseSrcSpanEndLine l) ++ ":"
+        ++ (show $ parseSrcSpanEndColumn l)]
+badExps _                                  = []
+
+badMatches :: Match SrcSpanInfo -> [String]
+badMatches (Match _ _ pats _ _) = [ "Bad pattern match!" | p <- pats, not (isGoodPatRhs p) ]
```
cdsmith: Fix the error message here, please.

codeworld-compiler/src/ParseCode.hs
```diff
@@ -0,0 +1,128 @@
...
+badMatches _                    = []
+
+isGoodExpRhs :: Exp l -> Bool
+isGoodExpRhs (Paren _ _)   = True
+isGoodExpRhs (Tuple _ _ _) = True
+isGoodExpRhs _             = False
+
+isGoodInfixExpRhs :: QOp l -> Bool
```
cdsmith: Seems to be dead code.  Please delete.

codeworld-compiler/src/ParseCode.hs
```diff
@@ -0,0 +1,128 @@
...
+isGoodInfixExpRhs (QVarOp _ _) = True
+isGoodInfixExpRhs _            = False
+
+isGoodPatRhs :: Pat l -> Bool
+isGoodPatRhs (PParen _ _)   = True
+isGoodPatRhs (PTuple _ _ _) = True
+isGoodPatRhs _              = False
+
+parseSrcSpanEndLine :: SrcSpanInfo -> Int
```
cdsmith: From here on down looks unnecessary.  It's obsolete because of other suggested changes above.

codeworld-compiler/src/ParseCode.hs
pranjaltale16: Should I update this like this?
```
program.hs:12:5: error: Bad Pattern Match
```
or it has to be something else?
@cdsmith

cdsmith: For now, you could use the same error as the expression case, actually.  Or you could say "constructor" instead of "function".  Either way is fine.

pranjaltale16:
I hope the error message in `badMatches` is correct.

pranjaltale16:
I updated the error messages of all the test-cases in order to match with haskell-src-exts error messages and also fix #541.

codeworld-compiler/src/ParseCode.hs
```diff
@@ -0,0 +1,64 @@
...
+            [] -> return True
+            errors -> writeFile err (intercalate "\n\n" errors) >> return False
+        ParseFailed _ _ -> return True
+
+getErrors :: Module SrcSpanInfo -> [String]
+getErrors m = everything (++) (mkQ [] badExps) m ++
+              everything (++) (mkQ [] badMatches) m
+
+badExps :: Exp SrcSpanInfo -> [String]
+badExps (App x (Var l _) e) | not (isGoodExpRhs e) = [
```
cdsmith: You can't limit to a Var on the left side of an application.  At the very least, you need to match Con, Lambda, LeftSection, RightSection... but why not just match anything, and use the location from the App?

pranjaltale16: If we take the starting point from the location of the App, then it will show, missing parentheses at the starting of the function name rather than at the end of it, which might confuse the user. One thing we can do is, keep this and add one more line like this:
```
badExps (App x _ e) | not (isGoodExpRhs e) = [
```
for other cases.

What say?

cdsmith: Actually, it will do that anyway, since formatLocation uses the start of the span.  Anyway, it's showing the starting point of the function application that is missing the parentheses, which seems logical to me.

Maybe there's some tweaking to be done, but I would rather be correct first, and tweak later, rather than start out with a bunch of kludges.

Issue #533, opened 2017-07-30
Inspect doesn't work well with coordinatePlane


cdsmith:
If you look at a coordinate plan (https://code.world/#PKt-AhweHztaNz5liE8X7kw), Inspect doesn't let you click anything there.  I guess that's because the source locations for the primitive shapes are not located in the main module?

cdsmith:
@three

Issue #531, opened 2017-07-30
Selection in inspect mode


cdsmith:
Just an idea, but it would be excellent if, in Inspect mode, after clicking a shape, one could navigate through the picture tree, by binding keys like:

Up - move up a level
Down - move to first child
Left - move to previous child, if any
Right - move to next child, if any

The assumption here is that we view the picture as a tree, where leaves are the primitive shapes, Pictures nodes have multiple children, and transformation nodes have a single child.  As one navigates through the tree, the selection in the editor can be changed to match.

cdsmith:
@three 

If #530 is implemented anyway, a highlight for a selection as well would be a great addition to this feature.

Issue #530, opened 2017-07-30
Mouse hover in Inspect mode


cdsmith:
The new inspect mode could be made much more usable if it reacted to mouseover by drawing a shaded selection region over the leaf shape beneath the cursor.

PR #549, opened 2017-09-03
Add Picture Dialog to browse picture tree

three:
When in debug mode, this PR replaces the infobox that would previously show up with a dialog showing more detailed and readable information about a picture. Additionally, mousing over a shape or shape name (in picture browser) will cause that shape to be highlighted. This also adds user-defined symbols to the bottom of autocomplete (see commit c63fd5c).

This PR makes a lot of changes to `Driver.hs`, namely an intermediate structure `Drawing` is added which makes highlighting possible.

A demo should be available at http://45.32.1.97/.

Even though the Summer of Haskell working period has ended, I'm still committed to maintaining my contributions to Codeworld. If there's any problems with this PR or bugs that show up in the future I'm still happy to fix them.


three:
This PR should fix #533, #530 and make #531 obsolete (or make fixing #531 a matter of adding keybindings).

cdsmith:
Thanks for the pull request.

I like the ideas here, but I don't think I can merge it in its current form due to UI issues.  It comes down to a number of small things, but put together, the interface is just too clunky and confusing.

A few of these small things are:

1. The code draws a popup window that looks almost taken from a 1987-era early GUI, with black boxes on a white background.  The title text is gigantic and out of proportion to the rest of the UI.  We don't need anything fancy here, but this just doesn't fit the look and feel of the web site at all, and it's confusing.
2. The slash-separated picture "path" isn't something that will likely make a lot of sense.
3. It's not clear to me why the "tree" section of the dialog box repeats the shape name, and doesn't seem to look much like a tree.
4. The coloring on highlight on the hover is far too intrusive.  It's frustrating to try to select a shape, when I have to memorize its location before moving the mouse over the canvas, because it will be hidden as soon as I get near it.  Perhaps it might look nicer if the highlight were always white or black with 25% alpha?

I also noticed that the polygon coordinates are wrong!  I'm not at all sure how, but the x coordinates shown in the popup are different from those in the code.

cdsmith:
http://45.32.1.97/#PdWWq7nFoR-TxEgt9JQRRtw (the tie) is an example where polygon coordinates are wrong.

three:
Thanks for the feedback! These should be easy changes, and I'll update the test server as I commit. I just want to make sure I'm going in the right direction with 1-3.

1\. The dialog is just a Jquery UI dialog. I styled it myself so I didn't have to include jquery's full stylesheet as a dependency. If I use jquery's stylsheet it'll look a bit more modern. http://jqueryui.com/resources/demos/dialog/default.html

2 and 3. My goal for the dialog was to make it look something like a filebrowser where users could navigate through the tree-like structure of a Picture. The path is the shape's location in a picture's tree. If you click on something up the path the tree will show a nested list of everything below it. Originally the picture browser only opened after clicking a button on a row of the infobox, so I could possibly remove the path and revert to that.

4\. This is a good point. I'll try changing the highlight color to always be black with a low alpha.

The coordinate problem seems to be caused by my `num2str` method not working on negative numbers. Deceivingly it still manages to produce something that looks reasonable.

cdsmith:
Okay, I completely missed the tree structure working out once you have selected a parent node.  That's actually pretty cool now.  I still think the combination of both path and tree is overkill, though, and that many of my students won't really have the concept of a file path to work with in the first place, so the metaphor will be lost.  What about just showing the collapsible tree (starting from the root) and pre-expanding just the nodes on the path all the way from the root to the selected node?

If that's done, the path will be gone.  The title bar of the popup was never useful.  The current selected shape name could be better indicated by visibly highlighting it in the tree, rather than repeating the name.

Then the question is what to do with the text description.  Initially, my thoughts are that if you can already select the relevant code (at least you used to be able to... I can't figure out how any more), then that information is already accessible that way.  It's true that the code doesn't label the meanings of the parameters as clearly, but I suspect that in the long run, the right solution to that will be to add the info in the editor instead of the debugger - e.g., in a tooltip that appears when the cursor is on a function application, as in http://lh5.ggpht.com/_0j4bzarlOBg/S3FQy-9UZDI/AAAAAAAAArM/hjVu0Y5xrXg/image_thumb%5B9%5D.png?imgmax=800

So I guess I wonder if the best approach is to *only* show the collapsible tree, render the selection there, and ALSO render the same selection in the code.

On the other hand, here's what would be absolutely lovely: if, when a node is selected, you also see just that one picture drawn onto a smaller canvas with a coordinate plane.  That would be incredibly useful for helping students to build a compositional mental model for pictures, and understanding that the subexpressions describe complete pictures in their own right.  So those are the two things that I think pass muster here: the collapsible tree, and a mini-canvas showing the selected picture on its own with a coordinate plane.

cdsmith:
As far as the UI style, there are a few things that seem less than ideal to me.  The easier problem is just that it looks flat, while the rest of the site doesn't.  You can probably go a long way just by adding the "dropbox" class to the window.

But I feel like modeless dialogs are just not something that you see any more in modern UI design, so it will feel foreign and strange.  I preferred the older approach with a temporary box -- but maybe it won't be as convenient when there's more to navigate through and you can't move the box out of the way.  (Another approach would be to replace or push down the message area; but the message area already gets chopped off on small screens, so that's probably not feasible.)

cdsmith:
Eric, thanks for the fixes to numbers and highlight color.  Are you still working on other concerns, or did you need anything from me to proceed?

three:
I'm working on drawing the selected picture to a smaller canvas. Sorry I'm a bit slow, class started for me this week and I had to worry about some school-related stuff.

I do have a question though. I think the tree would be difficult to navigate in the same box from before, even with most things collapsed, so I'm not quite sure where to go with the UI. For some pictures the tree can be quite large and deeply-nested. I could scrap the tree and go back to the stack list, with an extra button to show the picture with a coordinate plane. I don't see a good way to do the tree without covering up something else or using a dialog. Possibly a modal dialog with just the tree could be shown, but that loses some of it's interactivity.

It'll be easier to implement if I know what I'm aiming for, so I'm just looking for a direction to go in.

cdsmith:
Okay, for now let's keep the modeless dialog.  But applying the default styles (though I like your close button better) and adding the "dropbox" class, which adds a drop shadow for consistency with the rest of the UI would be a great change.  Then I'd love to see the full tree all the way from the root, and for clicking on a tree item to select the relevant code and possibly draw that shape, rather than changing the dialog contents.  Do you think that would work?

three:
The preview should be running in the dialog now (and live at http://45.32.1.97/). Hopefully I can add default styles and remove the path and description as well tonight. I'm not sure I can keep the current close button, although I've noticed hover doesn't show properly on Firefox which might be why jquery does it differently.

I think selecting the code and not changing dialog contents is a good idea. Should the picture show up on a different column of the dialog or by the mouse so students can still see it as their browsing the tree?

cdsmith:
This looks much nicer, I think.  I've been swamped with other things, but I'll try to take another detailed look in the next day or so.

cdsmith:
Cool!  There's a lot of stuff I'd love to try, but this definitely looks better than what is live now, so I'll merge and push other changes forward.

PR #551, opened 2017-09-08
Introduce Commenting Feature and Simultaneous Coding Environment.

parvmor:
Sorry for being so late in sending the PR. Some of the key highlights of the changes made are:

_**Note:**_ Please do merge this right now. I would post a script here to migrate existing users files to the new file structure. Or else if you want users to be migrated online as and when required I can make the required changes.

1. Splits up the server code into multiple modules(like Folder, Comment, Collaboration etc) and removes several avoidable GHC warnings.
2. The file structure of storing the user data is changed completely so as to accommodate multiple users to code together by referencing a single file and simultaneously allowing users to comment on each others code. (More on this later)
3. The save as button is removed and instead of that copy button is introduced which does the same job as save as but also handles folders.
4. Users will now not be able to type straight out into the editor. They would always be required to create a file whenever they want to code.
5. Users will be required to add a user identifier with each project(can vary with projects) that would appear in the comments. (_**Question:**_ What to do about existing files? Possible Solution: Online migration instead of offline)

## The new file structure:

- For each user we maintain a separate folder, namely `commentables`, which contains all the projects user can comment on but cannot edit. Note that user will not be able to make changes so in such a case I have a introduced a `Test Code` button which will generate a test environment.
- One can copy the projects/folders present in `commentables` into the editable category of project. Such copying will only preserve source codes. One cannot copy in to `commentables` directory.
- One can move the projects/folders present in `commentables` into `commentables` and not outside of that.
- To make multiple users to share a common file I introduced the following structure:
   1. Store all of the projects in `data/mode/projectContents` independent of which user they belong to. The project would be stored with a filename which is a hash of `data/mode/projects/userId/userPath`. Here userId will be chosen randomly from one of the many users that can edit this file. 
   2. To link the users to this project we store the path of that file at user's path and also the user's path at the common location. We can visualize this as a star graph with undirected edges.
   3. All the data of comments will be stored at this new common path and any new user who is added only needs to be associated with this location.

## Simultaneous coding:

- By default every user will always be in a collaborative coding mode. To collaborate share the url generated by `Collaborate` button and whoever accesses this link will be given complete access to edit.
- Deleting self from a collaborated will not delete it for others. Only the link between the user and project will be deleted.
- One can see current and active owners by clicking `Current Owners` button.

## Commenting:

- One can comment on a project by clicking on any line in the gutter of the editor.
- Handling line comments becomes a bit complicated when the file is changed. After discussing with Michael, I started doing versioning whenever needed.
- Previous versions will not be editable but can be accessed and tested out whenever required.
- Current commenting system is not interactive in the sense that it does not gives real time updates. But since the backend is set up it should be easy to integrate that using sockets. I would do that in my free time.

**_Note:_** For funblocks I have created a new server as it had conflicting handlers due to different file structures.

If I have missed anything new I will comment it below. Suggesstions are welcome. If any problems occur with this PR I will be glad to fix them.


codeworld-server/src/Collaboration_.hs
```diff
@@ -0,0 +1,246 @@
...
+    , ("listCurrentOwners", listCurrentOwnersHandler clientId)
+    , ("socket.io", socketIOHandler)
+    ]
+
+getFrequentParams :: Int -> ClientId -> Snap (User, BuildMode, FilePath)
+getFrequentParams getType clientId = do
+    user <- getUser clientId
+    mode <- getBuildMode
+    case getType of
+        1 -> do
```
cdsmith: Please, no!  At least, use the type system.  This is Haskell.  But better yet, decompose the problem in a way that's less arbitrary.

codeworld-server/src/Collaboration_.hs
```diff
@@ -0,0 +1,246 @@
...
+            Just path' <- fmap (splitDirectories . BC.unpack) <$> getParam "path"
+            Just name <- getParam "name"
+            let projectId = nameToProjectId $ T.decodeUtf8 name
+                finalDir = joinPath $ map (dirBase . nameToDirId . T.pack) path'
+                file = userProjectDir mode (userId user) </> finalDir </> projectFile projectId
+            case length path' of
+                0 -> return (user, mode, file)
+                _ -> case path' !! 0 of
```
cdsmith: Why not capture the first element in the top-level case?

cdsmith:
Parv, I started to take a look, but before I get much further, let me address a few high-level questions:

1. The changes to data/X/projects sound fine, but we will need a migration scripts.
2. Do you have a running demo server somewhere?  It would help tremendously to see the UI changes and be able to play around with them.  Otherwise, I can try to set up a second server myself.

funblocks-server/src/Funblocks.hs
```diff
@@ -0,0 +1,261 @@
...
+                when (clientId /= ClientId (Just (audience user))) pass
+                return user
+
+getBuildMode :: Snap BuildMode
+getBuildMode = getParam "mode" >>= \ case
+    Just "blocklyXML" -> return (BuildMode "blocklyXML")
+
+funblockRoutes :: ClientId -> [(B.ByteString, Snap ())]
+funblockRoutes clientId =
+    [ ("floadProject",   loadProjectHandler clientId)
```
cdsmith: This is extremely unfortunate.  Why should all the blocks UI stuff be so separate?

I feel like if I merge this change, that will be the day that Blocks starts to bitrot.  All of a sudden, it will share almost no code with the main site, and so its code will get further behind over time, until it's an unmaintainable mess.  I am not ready to give up on it yet.

third_party/ot.hs/.ghci
```diff
@@ -0,0 +1,4 @@
+:set -fwarn-unused-binds -fwarn-unused-imports
```
cdsmith: I cannot review your changes to the ot library, because I have no way to tell what's yours, and what is part of the original.  And, sorry, I can't do a complete code review for the third party ot library.

Is it possible to see a diff for your changes somehow?  Have you submitted your changes upstream, so that CodeWorld isn't maintaining a fork of this library indefinitely?

build.sh
```diff
@@ -43,9 +43,11 @@ run codeworld-api   cabal haddock --hoogle
 
 # Build codeworld-server from this project.
 
-run .  cabal_install ./codeworld-server \
+run .  cabal_install ./third_party/ot.hs \
```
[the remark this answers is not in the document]
parvmor: Wait which file? That is a directory. It contains the source code for operational transformations.

codeworld-server/codeworld-server.cabal
```diff
@@ -24,20 +24,30 @@ Executable codeworld-server
     cryptonite,
     data-default,
     directory,
+    engine-io,
+    engine-io-snap,
     filepath,
     filesystem-trees,
+    funblocks-server,
```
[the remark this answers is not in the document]
parvmor: The file structure for funblocks and codeworld are different. So, the handlers for handling files in both have different logic. I thought it would be best to have a separate library for the same. But I am open up for suggestions

codeworld-server/src/Collaboration.hs
```diff
@@ -0,0 +1,19 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-}
+
+module Collaboration (module Collaboration__) where
+
+import Collaboration_ as Collaboration__ hiding (getFrequentParams)
```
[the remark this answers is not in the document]
parvmor: This was used as a way to get around the fact that haskell requires you to explicitly export functions you want to export. This allows to hide the functions you do not want to export. But now I can see that looks kind of a hack and ugly. So, I will change that.

funblocks-server/src/Funblocks.hs
parvmor: I agree with you on this but new file structure for collaboration would require this to happen. A solution would be to let even funblocks use that but then it will only create some unnecessary files. Any suggestions for this?

third_party/ot.hs/.ghci
parvmor: Sorry for this. There were no changes made by me over here. I will remove this and add a dynamic installation of ot.hs.

parvmor:
As I mentioned the migration script we cannot(or preferable not to) add an offline migration script because user identifiers are need to be associated which each file. So, one way would be ask a common user identifier once and for all or to prompt for a user identifier whenever a file that doesn't has a user identifier associated with it. Which one do you prefer?

Also, I do not have a demo server that you would be able to access. I agree you should test them before merging. As of right now I do not have access to any server on which I can host. Can you try to setup one?(I will see if I can arrange one if possible. Having exams next week)

codeworld-server/codeworld-server.cabal
```diff
@@ -24,20 +24,30 @@ Executable codeworld-server
     cryptonite,
     data-default,
     directory,
+    engine-io,
```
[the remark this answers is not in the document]
parvmor: Sure would make the required changes!

codeworld-server/src/Collaboration_.hs
parvmor: Actually I don't why I didn't do that. Lazy evaluation should allow that. Will update that.

third_party/ot.hs/.ghci
parvmor: Wait as a matter of fact I did made changes to the cabal file. The dependencies specified were not being resolved.

```
Resolving dependencies...
cabal: Could not resolve dependencies:
trying: ot-0.2.1.0 (user goal)
next goal: ghc (dependency of ot-0.2.1.0)
rejecting: ghc-8.0.1/installed-8.0... (conflict: ghc =>
binary==0.8.3.0/installed-0.8..., ot => binary>=0.5.1.1 && <0.8)
Dependency tree exhaustively searched.
```
Resolving this still gave more dependency errors. So I removed the versions. The only change is the following:

```
25,32c25,32
<   Build-depends:       base,
<                        text,
<                        aeson,
<                        attoparsec,
<                        QuickCheck,
<                        binary,
<                        either,
<                        mtl,
---
>   Build-depends:       base >= 4 && < 5,
>                        text >= 1.0 && < 1.3,
>                        aeson >= 0.7 && < 0.11,
>                        attoparsec >= 0.10.1.1 && < 1,
>                        QuickCheck >= 2.7 && < 2.9,
>                        binary >= 0.5.1.1 && < 0.9,
>                        either >= 4.1.2 && < 5,
>                        mtl >= 2.1.3.1 && < 3,
```

funblocks-server/src/Funblocks.hs
cdsmith: It seems like most of the file manipulation should be the same.  The blocks UI should have no trouble supporting folders and such.  The only change should be disabling collaborative editing.  (In theory, that could eventually be implemented too; it would just require a different OT model and integration with Blockly!)  My hope would be that they would mostly share code, with a few buttons being disabled.  But I don't understand the functionality yet, so it's hard for me to propose a specific answer.

cdsmith:
I still need to set up a test server and try this out.  Unfortunately, the last week has been very busy for me.  I'll try to get to this ASAP.

PR #589, opened 2018-01-26
Added content to introduction

chelseaharper:


web/help/GuideIntro.md
```diff
@@ -15,6 +15,15 @@ back here.  The outline to the left will allow you to jump between sections of t
 Overview
 ========
 
+If you're in a middle school class taking a course using Codeworld, this guide should
```
cdsmith: Just reiterating what we talked about this morning: even though the online guide follows roughly the same outline as the curriculum, let's keep the two independent of each other.  The guidance to students on how to match up sections should all go on the curriculum side, so someone working with the online guide need not feel that anything is missing.

cdsmith:
This became obsolete.  I'll manually incorporate the comment about working from scratch, which is a good one.

PR #599, opened 2018-03-16
Allow building on OS X

tgdavies:
This builds and runs on my machine -- it's possible that I already had some dependencies installed, so it may or may not work on a clean OS X install.

It requires that Xcode and homebrew are installed.


tgdavies:
Oops -- I've committed with my work email address -- I'll redo the PR :-(

PR #600, opened 2018-03-24
Add OS X package manager and GHC architecture detection

tgdavies:
This works on my machine, and it's possible that I have some dependencies already installed.

It runs OK, but when I run the logo demo I get the error messages:

Line 30, Column 48: error: Missing parentheses in function application
Line 42, Column 31: error: Missing parentheses in function application
Line 43, Column 31: error: Missing parentheses in function application

even though the logo seems to be produced correctly.
screen shot 2018-03-24 at 10 32 47 pm

### DIFF
--- a/base.sh
+++ b/base.sh
@@ -36,5 +36,5 @@ function run {
 }
 
 function cabal_install {
-  cabal install --force-reinstalls --global --prefix=$BUILD $@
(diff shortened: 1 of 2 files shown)


cdsmith:
Those messages are expected, as a consequence of changes made over the summer.  I'll fix the demo.

Thanks!

PR #612, opened 2018-05-02
Update install.sh

kammitama5:
updated for i386 GHC instructions

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -167,15 +167,16 @@ case `uname -m` in
   i686)   GHC_CPU=i386;;
   x86_64) GHC_CPU=x86_64;;
   amd7)   GHC_CPU=amd7;;
+  *) >&2 echo "Unrecognized machine: ${MACHINE}"; exit 1;;
 esac
 
 ## then
 if /sbin/ldconfig -p | grep -q libgmp.so.10; then
-  GHC_ARCH=$GHC_CPU-deb8-linux
+  GHC_ARCH="${GHC_CPU}-deb8-linux"
 elif /sbin/ldconfig -p | grep -q libgmp.so.3; then
-  GHC_ARCH=$GHC_CPU-centos67-linux
+  GHC_ARCH="${GHC_CPU}-centos67-linux"
 elif uname | grep -q Darwin; then
-  GHC_ARCH=$GHC_CPU-apple-darwin
+  GHC_ARCH="${GHC_CPU}-apple-darwin"
 else
   echo Sorry, but no supported libgmp is installed.
   exit 1
--- a/install.sh
+++ b/install.sh
@@ -163,7 +163,7 @@ fi
 
 # Choose the right GHC download
 MACHINE="$(uname -m)"
-case `uname -m` in
+case "${MACHINE}" in
   i386)   GHC_CPU=i386;;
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -162,12 +162,20 @@ else
 fi
 
 # Choose the right GHC download
+case `uname -m` in
+  i386)   GHC_CPU=i386;;
+  i686)   GHC_CPU=i386;;
+  x86_64) GHC_CPU=x86_64;;
+  amd7)   GHC_CPU=amd7;;
```
Gabriel439: I suggest also adding a fallback case that fails if this runs into an unexpected platform type, i.e.:

```bash
MACHINE="$(uname -m)"
case "${MACHINE}" in
    ...
    *) >&2 echo "Unrecognized machine: ${MACHINE}"; exit 1;;
```

install.sh
```diff
@@ -162,12 +162,20 @@ else
...
+esac
+
+## then
 if /sbin/ldconfig -p | grep -q libgmp.so.10; then
-  GHC_ARCH=`uname -m`-deb8-linux
+  GHC_ARCH=$GHC_CPU-deb8-linux
```
Gabriel439: It's not really necessary here, but I recommend quoting shell variable interpolation as a habit to avoid all sorts of nasty shell gotchas:

```bash
GHC_ARCH="${GHC_CPU}-deb8-linux"
```

install.sh
kammitama5: something along these lines? Not that familiar with shell syntax so was unsure whether the quotes were necessary

cdsmith: You've used an undefined variable `MACHINE`.  You can either:

1. Define `MACHINE` at the beginning, as in Gabriel's example, OR
2. Use `$(uname -m)` instead of `$MACHINE`.

install.sh
```diff
@@ -162,12 +162,21 @@ else
...
 # Choose the right GHC download
+case `uname -m` in
+  i386)   GHC_CPU=i386;;
+  i686)   GHC_CPU=i386;;
+  x86_64) GHC_CPU=x86_64;;
+  amd7)   GHC_CPU=amd7;;
+  *) >&2 echo "Unrecognized machine: ${MACHINE}"; exit 1;;
+esac
+
+## then
```
cdsmith: Stray comment?  I'm not sure what it meant.

kammitama5: the ```## then```? Thought it was for originally,  go through cases ```##then``` choose appropriate download

kammitama5: Oh, I just re-pasted the formatting so that it matched what I had previously. It was previously there..didn't actually change anything.

install.sh
cdsmith: Since you defined `$MACHINE`, can you use it in line 166?  So that would say `case "${MACHINE}" in`.  There's no need to both define a variable and rerun the `uname` command.

install.sh
cdsmith: Yes, I meant `## then`.  It is not there before this change, so I suppose it just snuck in by accident.  Can you remove it, please?

install.sh
kammitama5: Thanks. Just corrected it and removed stray comment.

PR #618, opened 2018-05-09
Fix bad markdown formatting

byorgey:
"Radius of\n2.  Blah..." was apparently getting parsed by Markdown as the start of an enumeration.


cdsmith:
Thanks

PR #619, opened 2018-05-14
Clean install

kammitama5:


codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -1387,6 +1387,7 @@ getWebSocketURL = do
             case proto of
                 "http:" -> "ws://" <> hostname <> ":9160/gameserver"
                 "https:" -> "wss://" <> hostname <> "/gameserver"
+                _-> error "Unrecognized protocol"
```
cdsmith: Just a nit, but can you please add a space after the _ to keep the same style?

kammitama5:
Okey doke. Just did.

PR #624, opened 2018-05-16
Add SolidClosedCurve

kammitama5:
```solidPolygon``` function still uses ```SolidPolygon``` constructor and ```solidClosedCurve``` uses SolidClosedCurve. It type-checks, but not sure if I did anything wrong ;/

### DIFF
--- a/codeworld-api/src/CodeWorld/Driver.hs
+++ b/codeworld-api/src/CodeWorld/Driver.hs
@@ -210,8 +210,8 @@ type DrawState = (Double, Double, Double, Double, Double, Double, Maybe Color)
 type NodeId = Int
 
 pictureToDrawing :: Picture -> Drawing
-pictureToDrawing (SolidClosedCurve _ pts s) = Shape $ polygonDrawer pts True
-pictureToDrawing (SolidPolygon _ pts s) = Shape $ polygonDrawer pts False
+pictureToDrawing (SolidClosedCurve _ pts ) = Shape $ polygonDrawer pts True
+pictureToDrawing (SolidPolygon _ pts ) = Shape $ polygonDrawer pts False
 pictureToDrawing (Path _ pts w c s) = Shape $ pathDrawer pts w c s
 pictureToDrawing (Sector _ b e r) = Shape $ sectorDrawer b e r
 pictureToDrawing (Arc _ b e r w) = Shape $ arcDrawer b e r w
@@ -439,15 +439,15 @@ picToObj = fmap fst . flip State.runStateT 0 . picToObj'
 picToObj' :: Picture -> State.StateT Int IO JSVal
 picToObj' pic =
     case pic of
-        SolidPolygon cs pts smooth -> do
+        SolidPolygon cs pts -> do
             obj <- init "solidPolygon"
             ptsJS <- pointsToArr pts
-            setProps [("points", ptsJS), ("smooth", pToJSVal smooth)] obj
+            setProps [("points", ptsJS), ("smooth", pToJSVal True)] obj
             retVal obj
-        SolidClosedCurve cs pts smooth -> do
+        SolidClosedCurve cs pts -> do
             obj <- init "solidClosedCurve"
             ptsJS <- pointsToArr pts
-            setProps [("points", ptsJS), ("smooth", pToJSVal smooth)] obj
+            setProps [("points", ptsJS), ("smooth", pToJSVal False)] obj
             retVal obj
         Path cs pts w closed smooth -> do
             obj <- init "path"
@@ -573,8 +573,8 @@ findCSMain cs =
     Data.List.find ((== "main") . srcLocPackage . snd) (getCallStack cs)
 
 getPictureCS :: Picture -> CallStack
-getPictureCS (SolidPolygon cs _ _) = cs
-getPictureCS (SolidClosedCurve cs _ _) = cs
+getPictureCS (SolidPolygon cs _ ) = cs
+getPictureCS (SolidClosedCurve cs _ ) = cs
 getPictureCS (Path cs _ _ _ _) = cs
 getPictureCS (Sector cs _ _ _) = cs
 getPictureCS (Arc cs _ _ _ _) = cs
--- a/codeworld-api/src/CodeWorld/Driver.hs
+++ b/codeworld-api/src/CodeWorld/Driver.hs
@@ -210,8 +210,8 @@ type DrawState = (Double, Double, Double, Double, Double, Double, Maybe Color)
 type NodeId = Int
 
 pictureToDrawing :: Picture -> Drawing
-pictureToDrawing (SolidClosedCurve _ pts s) = Shape $ polygonDrawer pts s
-pictureToDrawing (SolidPolygon _ pts s) = Shape $ polygonDrawer pts s
+pictureToDrawing (SolidClosedCurve _ pts s) = Shape $ polygonDrawer pts True
+pictureToDrawing (SolidPolygon _ pts s) = Shape $ polygonDrawer pts False
 pictureToDrawing (Path _ pts w c s) = Shape $ pathDrawer pts w c s
 pictureToDrawing (Sector _ b e r) = Shape $ sectorDrawer b e r
 pictureToDrawing (Arc _ b e r w) = Shape $ arcDrawer b e r w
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: codeworld-api/src/CodeWorld/Picture.hs [hunk 1]
```diff
@@ -51,6 +51,9 @@ data Picture
     = SolidPolygon CallStack
               [Point]
               !Bool
+    | SolidClosedCurve CallStack
+              [Point]
+              !Bool
```
cdsmith: You should no longer need this `Bool`.  It only existed to distinguish between polygons and closed curves (i.e., smooth or not).  Since they are separate constructors, you no longer need to keep track.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -443,6 +444,11 @@ picToObj' pic =
             ptsJS <- pointsToArr pts
             setProps [("points", ptsJS), ("smooth", pToJSVal smooth)] obj
             retVal obj
+        SolidClosedCurve cs pts smooth -> do
+            obj <- init "SolidClosedCurve"
```
cdsmith: Let's stick to the exported function name?  So `solidClosedCurve` rather than `SolidClosedCurve`.

codeworld-api/src/CodeWorld/Picture.hs
kammitama5: Tricky stuff. May need some help on this one :D I pushed the small edit up in the meantime (ie "S" to "s")

cdsmith: Sure.  The way you've got it now, `SolidPolygon` always carries a value of `False` for the last field, and `SolidClosedCurve` always carries a value of `True`.  You can just remove the field from the type, and when you're pattern matching, you'll know whether it should be `True` or `False` based on which constructor it is.

kammitama5: I removed the ```Bool``` constructors from ```SolidPolygon``` and ```SolidClosedCurve``` and then I get this. I think I know what to do with the first error, but the case statement looks like it's a bigger fix and I'm not sure if I'm on the right track seeing this.

screenshot_2018-05-15_23-52-53

kammitama5: How do I resolve this type? I figured that ```s``` is ```smooth``` (so True or False) and was able to resolve the other parts by removing the ```smooth parameter``` in the case statements, but some guidance on this last part would be really appreciated :D 

screenshot_2018-05-16_00-55-38

cdsmith: Sounds good

kammitama5:
Gabe suggested that I continue to make PRs when they type-check. I'm adding this as I added the ```True```  and ```False``` explicitly in ```Driver.hs``` and it type-checks.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -210,7 +210,8 @@ type DrawState = (Double, Double, Double, Double, Double, Double, Maybe Color)
 type NodeId = Int
 
 pictureToDrawing :: Picture -> Drawing
-pictureToDrawing (SolidPolygon _ pts s) = Shape $ polygonDrawer pts s
+pictureToDrawing (SolidClosedCurve _ pts s) = Shape $ polygonDrawer pts True
```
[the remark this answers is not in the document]
kammitama5: but then it gives me errors that I need to remove the Bool fields of the ```solidPolygon```, ```solidClosedCurve``` and ```solidClosedLoop``` functions (lines 138, 171, 175) in Pictures.hs, and once I resolve those, I've been having difficulty trying to get type-checking in Driver.hs because it keeps wanting me to change values in ```PictureToDrawing``` and ```picObj```. :/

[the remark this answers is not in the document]
Gabriel439: So getting rid of the `Bool` fields of `solidPolygon`/`solidClosedCurve` and `solidClosedLoop` is an expected outcome of that change.  What is the type error that you run into in `Driver.hs`?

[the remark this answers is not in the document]
kammitama5: I get three. One on line 213 in ```Driver.hs``` expecting two arguments (which is fixable), then one in line 442 wanting two arguments (but not sure I should dick with that), and one on 576 wanting two arguments. I've tried in the past to fix those and ended up with the error I got two days about ```polygonDrawer``` (above image error)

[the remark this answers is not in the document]
Gabriel439: So the errors in `picToObj` you should fix in the same way that you fixed `pictureToDrawing`, by supplying the correct `Bool` based on whether you are drawing a `SolidPolygon` or `SolidClosedCurve`

For the error in `getPictureCS`, the correct thing to do is to delete the unused argument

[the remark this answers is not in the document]
kammitama5: Thank you!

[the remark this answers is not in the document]
Gabriel439: You're welcome!

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -210,7 +210,8 @@ type DrawState = (Double, Double, Double, Double, Double, Double, Maybe Color)
 type NodeId = Int
 
 pictureToDrawing :: Picture -> Drawing
-pictureToDrawing (SolidPolygon _ pts s) = Shape $ polygonDrawer pts s
+pictureToDrawing (SolidClosedCurve _ pts ) = Shape $ polygonDrawer pts True
```
cdsmith: [nitpicking] No need for the space before close parenthesis.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -567,7 +573,8 @@ findCSMain cs =
     Data.List.find ((== "main") . srcLocPackage . snd) (getCallStack cs)
 
 getPictureCS :: Picture -> CallStack
-getPictureCS (SolidPolygon cs _ _) = cs
+getPictureCS (SolidPolygon cs _ ) = cs
```
cdsmith: More unnecessary spaces. :)

codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -167,11 +168,11 @@ thickLoop n ps = Path callStack ps n True True
 
 -- | A solid smooth closed curve passing through these points.
 solidClosedCurve :: HasCallStack => [Point] -> Picture
-solidClosedCurve ps = SolidPolygon callStack ps True
+solidClosedCurve ps = SolidClosedCurve callStack ps
 
 -- | A solid smooth closed curve passing through these points.
 solidLoop :: HasCallStack => [Point] -> Picture
-solidLoop ps = SolidPolygon callStack ps True
+solidLoop ps = SolidPolygon callStack ps
```
cdsmith: This should be a SolidClosedCurve.

kammitama5: Thank you! Just made the changes.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -438,10 +439,15 @@ picToObj = fmap fst . flip State.runStateT 0 . picToObj'
 picToObj' :: Picture -> State.StateT Int IO JSVal
 picToObj' pic =
     case pic of
-        SolidPolygon cs pts smooth -> do
+        SolidPolygon cs pts -> do
             obj <- init "solidPolygon"
             ptsJS <- pointsToArr pts
-            setProps [("points", ptsJS), ("smooth", pToJSVal smooth)] obj
+            setProps [("points", ptsJS), ("smooth", pToJSVal True)] obj
```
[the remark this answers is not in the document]
kammitama5: Ah yes. Sorry; I was at that point just getting it to type-check. I should have caught this.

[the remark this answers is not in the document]
cdsmith: No problem!  That's what code review is for.  I'll merge the pull request as soon as this is fixed.

cdsmith:
Merged.  Thanks!

PR #630, opened 2018-05-19
Start splitting Path into constructors

kammitama5:
it type-checked.

### DIFF
--- a/codeworld-api/src/CodeWorld/Driver.hs
+++ b/codeworld-api/src/CodeWorld/Driver.hs
@@ -461,7 +461,7 @@ picToObj' pic =
             ptsJS <- pointsToArr pts
             setProps
                 [ ("points", ptsJS)
-                , ("width", pToJSVal (0::Double))
+                , ("width", pToJSVal (0 :: Double))
                 , ("closed", pToJSVal True)
                 , ("smooth", pToJSVal False)
                 ]
@@ -483,7 +483,7 @@ picToObj' pic =
             ptsJS <- pointsToArr pts
             setProps
                 [ ("points", ptsJS)
-                , ("width", pToJSVal (0::Double))
+                , ("width", pToJSVal (0 :: Double))
                 , ("closed", pToJSVal True)
                 , ("smooth", pToJSVal True)
                 ]
@@ -505,7 +505,7 @@ picToObj' pic =
             ptsJS <- pointsToArr pts
             setProps
                 [ ("points", ptsJS)
-                , ("width", pToJSVal (0::Double))
+                , ("width", pToJSVal (0 :: Double))
                 , ("closed", pToJSVal False)
                 , ("smooth", pToJSVal False)
                 ]
@@ -523,11 +523,11 @@ picToObj' pic =
                 obj
             retVal obj
         Curve cs pts -> do
-            obj <- init "Curve"
+            obj <- init "curve"
             ptsJS <- pointsToArr pts
             setProps
                 [ ("points", ptsJS)
-                , ("width", pToJSVal (0::Double))
+                , ("width", pToJSVal (0 :: Double))
                 , ("closed", pToJSVal False)
                 , ("smooth", pToJSVal True)
                 ]
--- a/codeworld-api/src/CodeWorld/Driver.hs
+++ b/codeworld-api/src/CodeWorld/Driver.hs
@@ -213,13 +213,13 @@ pictureToDrawing :: Picture -> Drawing
 pictureToDrawing (SolidClosedCurve _ pts) = Shape $ polygonDrawer pts True
 pictureToDrawing (SolidPolygon _ pts) = Shape $ polygonDrawer pts False
 pictureToDrawing (Polygon _ pts) = Shape $ pathDrawer pts 0 True False
-pictureToDrawing (ThickPolygon _ pts) = Shape $ pathDrawer pts w True False
+pictureToDrawing (ThickPolygon _ pts w) = Shape $ pathDrawer pts w True False
 pictureToDrawing (ClosedCurve _ pts) = Shape $ pathDrawer pts 0 True True
-pictureToDrawing (ThickClosedCurve _ pts) = Shape $ pathDrawer pts w True True
+pictureToDrawing (ThickClosedCurve _ pts w) = Shape $ pathDrawer pts w True True
 pictureToDrawing (Polyline _ pts) = Shape $ pathDrawer pts 0 False False
-pictureToDrawing (ThickPolyline _ pts) = Shape $ pathDrawer pts w False False
+pictureToDrawing (ThickPolyline _ pts w) = Shape $ pathDrawer pts w False False
 pictureToDrawing (Curve _ pts) = Shape $ pathDrawer pts 0 False True
-pictureToDrawing (ThickCurve _ pts) = Shape $ pathDrawer pts w False True
+pictureToDrawing (ThickCurve _ pts w) = Shape $ pathDrawer pts w False True
 pictureToDrawing (Sector _ b e r) = Shape $ sectorDrawer b e r
 pictureToDrawing (Arc _ b e r w) = Shape $ arcDrawer b e r w
 pictureToDrawing (Text _ sty fnt txt) = Shape $ textDrawer sty fnt txt
@@ -467,7 +467,7 @@ picToObj' pic =
                 ]
                 obj
             retVal obj
-        ThickPolygon cs pts -> do
+        ThickPolygon cs pts w -> do
             obj <- init "thickPolygon"
             ptsJS <- pointsToArr pts
             setProps
@@ -489,7 +489,7 @@ picToObj' pic =
                 ]
                 obj
             retVal obj
-        ThickClosedCurve cs pts -> do
+        ThickClosedCurve cs pts w -> do
             obj <- init "thickClosedCurve"
             ptsJS <- pointsToArr pts
             setProps
@@ -500,7 +500,7 @@ picToObj' pic =
                 ]
                 obj
             retVal obj
-        Polyline cs pts-> do
+        Polyline cs pts -> do
             obj <- init "polyline"
             ptsJS <- pointsToArr pts
             setProps
@@ -511,7 +511,7 @@ picToObj' pic =
                 ]
                 obj
             retVal obj
-        ThickPolyline cs pts -> do
+        ThickPolyline cs pts w -> do
             obj <- init "thickPolyline"
             ptsJS <- pointsToArr pts
             setProps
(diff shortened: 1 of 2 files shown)


codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -460,6 +468,94 @@ picToObj' pic =
                 ]
                 obj
             retVal obj
+        Polygon cs pts w closed smooth -> do
+            obj <- init "path"
```
cdsmith: The names here can be changed to match the functions: `polygon`, `thickPolygon`, etc.

codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -57,6 +57,46 @@ data Picture
            !Double
```
cdsmith: You shouldn't need the `Path` costructor any longer.  `Path` is just an old name for `Polyline` (just like `loop` is an old name for `closedCurve`)

codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -57,6 +57,46 @@ data Picture
...
            !Bool
            !Bool
+    | Polygon CallStack
+           [Point]
+           !Double
```
cdsmith: Removing these extra parameters is part of the point of splitting this up.  You don't need to remember the width of a polygon, nor whether it's closed or smooth, because a polygon by definition has a width of 0, is closed, and is not smooth.  You should just need the `CallStack`, list of points, and ONLY for the thick variants, the thickness.

codeworld-api/src/CodeWorld/Picture.hs
kammitama5: But what about the path functions? What should I use in place of those (and functions such as ```path```, ```thickPath```, ```loop``` and ```thickLoop```)? Should I use ```closedCurve``` in their place?

cdsmith: `path` is the same as `polyline`
`thickPath` is `thickPolyline`
`loop` is `closedCurve`
`thickLoop` is `thickClosedCurve`

cdsmith: By the way, "path" and "loop" are deprecated.  They only exist because students used to use them a long time ago.  I'll probably drop support for them either this summer or next summer.

codeworld-api/src/CodeWorld/Picture.hs
kammitama5: silly question, but for ```Driver.hs```, do I need the n and w parameters (in ```pictureToDrawing```, etc)? They're both Doubles. I remember you had said something last week about having values (eg 1 through 4) specific to each type of Path value.

cdsmith: If I were you, I'd focus on removing them from the `Picture` type.  In `pictureToDrawing`, you can infer them from the `Picture` constructor.  For instance, if the constructor is `Curve`, then width is 0, smooth is true, and closed is false

kammitama5: I see. Thanks. I'll work on it this evening.

kammitama5: I've removed them but am still having problems resolving the type (specifically ```n``` and ```w```). I could push up what I have if that helps. I had tried to previously replace them with literals (1 through 4), but it didn't type-check. I understand that the n and w are Doubles and I've removed them from the ```Picture``` input types in ```picToObj'``` and ```pictureToDrawing```. So currently, I have something like ```pictureToDrawing (SolidPolygon _ pts) = Shape $ pathDrawer pts w True False``` and my w is not in scope because I deleted it from ```Picture.hs```. But it still needs a double. Similarly with the input parameters of say 
````Polygon cs pts -> do 
   obs <- init "polygon"
ptsJS <- pointsToArr pts 
setProps
[("points",ptsJS)
, ("width", pToJSVal w)
, ("closed", pToJSVal True)
, ("smooth", pToJSVal False)
]
obj
retVal obj
```
This is where I'm stuck. I'm trying to resolve the n/w Bool types for these, considering I've deleted them from the ```Picture``` arguments. Don't know if this makes sense.

cdsmith: I don't know what you mean by "the `n`".  The w parameter is still needed for thick shapes, but it is always zero (so not needed) for the others.

Please do push your current code when you have a question.  It helps a lot.

codeworld-api/src/CodeWorld/Picture.hs
cdsmith: Oh, I see now.  It looks like `n` is sometimes used instead of `w`.  They mean the same thing.

codeworld-api/src/CodeWorld/Picture.hs
cdsmith: I see now.  It looks like `n` is used instead of `w` sometimes, but they mean the same thing.

kammitama5: Just pushed it with the heading "this does not typecheck". I believe that ```Picture.hs``` is OK. I'm working on ```Driver.hs``` at the moment

codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -53,10 +53,23 @@ data Picture
     | SolidClosedCurve CallStack
               [Point]
     | Path CallStack
+              [Point]
+    | Polygon CallStack
+           [Point]
+    | ThickPolygon CallStack
+           [Point]
```
cdsmith: You still need a width for thickPolygon.  Same for all the other Thick variants.

kammitama5: this seems to work (dunno if I'm doing it right), but what do I do with the picObj that has w in scope in the return type but doesn't actually use w? I'll push up the code also as I have it.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -575,7 +659,14 @@ findCSMain cs =
 getPictureCS :: Picture -> CallStack
 getPictureCS (SolidPolygon cs _) = cs
 getPictureCS (SolidClosedCurve cs _) = cs
-getPictureCS (Path cs _ _ _ _) = cs
+getPictureCS (Polygon cs _) = cs
+getPictureCS (ThickPolygon cs _ w) = cs
```
cdsmith: You can use `_` instead of `w` here.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -449,14 +456,91 @@ picToObj' pic =
...
             setProps [("points", ptsJS), ("smooth", pToJSVal True)] obj
             retVal obj
-        Path cs pts w closed smooth -> do
-            obj <- init "path"
+        Polygon cs pts -> do
+            obj <- init "polygon"
             ptsJS <- pointsToArr pts
             setProps
                 [ ("points", ptsJS)
                 , ("width", pToJSVal w)
```
cdsmith: Did you mean 0?  You only have a `w` in scope for the thick variants?

codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -53,10 +53,27 @@ data Picture
     | SolidClosedCurve CallStack
               [Point]
     | Path CallStack
```
cdsmith: Why do you still have a `Path` here?  You aren't using it any more.

codeworld-api/src/CodeWorld/Picture.hs
cdsmith: The width can be set to zero for shapes that are not thick.  That will do the same thing it did before.

codeworld-api/src/CodeWorld/Driver.hs
kammitama5: I tried to pass w for the thick and 0 for the non-thick variants

cdsmith: This is Polygon, which is non-thick.  But you have a w here.  I suspect it won't build, since there is no w in scope.

kammitama5: Yes. That's what I was struggling to resolve. I tried putting a literal, but it won't accept literals. The Path found its way back in the constructor because I Ctrl+Z-ed a bit. I've removed it.

kammitama5: Okay. Got it to type-check.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -449,14 +456,91 @@ picToObj' pic =
...
             setProps [("points", ptsJS), ("smooth", pToJSVal True)] obj
             retVal obj
-        Path cs pts w closed smooth -> do
-            obj <- init "path"
+        Polygon cs pts -> do
+            obj <- init "polygon"
+            ptsJS <- pointsToArr pts
+            setProps
+                [ ("points", ptsJS)
+                , ("width", pToJSVal True)
```
cdsmith: Shouldn't this be 0 instead of True?  Same in 4 other places.

kammitama5: I couldn't get it to pass when I was passing 0. That was the w/n Double value I was trying to resolve. It type checked this morning when I passed a bool ie True.

cdsmith: Ah, you might need `0 :: Double` because of overloaded number literals and the weird type class.

kammitama5: I had tried that and kept getting this :/
screenshot_2018-05-20_18-42-54

cdsmith: You have a precedence problem.  The expression `pToJSVal 0 :: Double` is parsed as `(pToJSVal 0) :: Double` (in other words, it asserts that the output of the `pToJSVal` function is a `Double`), when you really meant `pToJSVal (0 :: Double)`.  You need the parentheses to get the latter meaning.

kammitama5: Ah. That's really cool! Thanks!

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -449,14 +456,91 @@ picToObj' pic =
...
             setProps [("points", ptsJS), ("smooth", pToJSVal True)] obj
             retVal obj
-        Path cs pts w closed smooth -> do
-            obj <- init "path"
+        Polygon cs pts -> do
+            obj <- init "polygon"
+            ptsJS <- pointsToArr pts
+            setProps
+                [ ("points", ptsJS)
+                , ("width", pToJSVal (0::Double))
```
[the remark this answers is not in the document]
kammitama5: Nitpicking is great. I appreciate it. Thank you!

cdsmith:
Thanks!  Merged.

PR #638, opened 2018-05-22
Splitting Arc and ThickArc in Picture - type checks first attempt

kammitama5:
What needs to be removed here? I've initially just passed ```0``` for Arc and Thickline has ```w``` and made it type-check.

### DIFF
--- a/codeworld-api/src/CodeWorld/Picture.hs
+++ b/codeworld-api/src/CodeWorld/Picture.hs
@@ -81,6 +81,11 @@ data Picture
           !Double
           !Double
           !Double
+    | ThickArc CallStack
+          !Double
+          !Double
+          !Double
+          !Double
     | Text CallStack
            !TextStyle
            !Font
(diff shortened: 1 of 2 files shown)


codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -81,6 +81,11 @@ data Picture
           !Double
           !Double
           !Double
```
cdsmith: I don't think you need this fourth parameter (the `w` or line width) to `Arc`.  It doesn't look like it's used anywhere.

kammitama5: I removed the fourth parameter from ```Arc```. The width is used for ```ThickArc```? Or should I remove that as well? :/

cdsmith: Line width is used for all thick shapes.

PR #640, opened 2018-05-22
TextStyled constructor (does not typecheck)

kammitama5:
I just started on the ```Text``` and ```StyledText``` and something is confusing me. I *think* ```Text``` should just take a ```Text``` argument in the constructor, and ```StyledText``` should take a ```TextStyle```, ```Font```, ```Text``` but I'm not sure. Is this along the lines that you were thinking? I can also work on this lunch time and tonight.

### DIFF
--- a/codeworld-api/src/CodeWorld/Driver.hs
+++ b/codeworld-api/src/CodeWorld/Driver.hs
@@ -586,7 +586,7 @@ picToObj' pic =
         StyledText cs style font txt -> do
             obj <- init "styledText"
             setProps
-                [ ("font", pToJSVal $ fontString Plain Serif)
+                [ ("font", pToJSVal $ fontString style font)
                 , ("text", pToJSVal txt)
                 ]
                 obj


codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -243,10 +245,10 @@ sector = Sector callStack
 
 -- | A piece of text
 text :: HasCallStack => Text -> Picture
-text = Text callStack Plain Serif
+text = Text callStack
 
 styledText :: HasCallStack => TextStyle -> Font -> Text -> Picture
-styledText = Text callStack
+styledText = StyledText callStack Plain Serif
```
cdsmith: I don't think you want Plain and Serif here.

Admittedly, this is a bit confusing.  There's some eta expansion needed to see all the parameters.  That comes from the fact that these lines of code:

    f = g
    f x = g x

mean the same thing.  Technically, the first line says that f and g are the same, while the second says that they have the same result for any input.  But since a function is *completely* determined by its result for each input, that's enough to conclude that they are the same anyway!  This fact is sometimes called "eta equivalence", and converting an equation `f = g` into the equation `f x = g x` is called "eta expansion".

The same thing was happening here.  The original code was `styledText = Text callStack`.  But after an eta expansion, that's the same thing as `styledText style = Text callStack style`, and two MORE eta expansions gives you `styledText style font txt = Text callStack style font txt`.  Now you can see all four arguments to the constructor.

The original style (`f = g` or `styledText = Text callStack`) is nice because it lets you talk about entire functions, instead of their specific values at specific parameters.  But I suppose it's not so nice if you aren't familiar with it.

kammitama5: Thank you so much for this!

kammitama5:
I specifically don't know how to resolve Text in Driver.hs both in PictureDrawing and in the JSVal. Sorry. Had to resolve a merge conflict :/

Gabriel439:
@kammitama5: The way you can resolve it is to examine what `style` and `font` would have been before you split `Text` into `Text`/`StyledText`

For example, before your change the `text` function was defined as:

```haskell
text = Text callStack Plain Serif
```

... but after the change it is defined as:

```haskell
text = Text callStack
```

... which means that anything using the new smaller `Text` constructor needs to have the `style` hard-coded to `Plain` and the `font` hard-coded to `Serif`.  In a sense, you are taking the `Plain` and `Serif` out of the definition of the `text` function and pushing `Plain`/`Serif` downstream to every place that pattern matches on `Text`.

So in the specific example of `Driver.hs`, in this code segment:

```haskell
        Text cs txt -> do
            obj <- init "text"
            setProps
                [ ("font", pToJSVal $ fontString style font)
                , ("text", pToJSVal txt)
                ]
                obj
            retVal obj
```

You can replace each occurrence of `font` in that block with `Serif` and each occurrence of `style` with `Plain`

Once you complete that, you could also go a step further and notice that the pattern match on the `Text` constructor has basically the same logic as the pattern match on the `StyledText` constructor, so you could factor the shared logic into a function, like this:

```haskell
  renderStyledText cs style font txt = do
            obj <- init "text"
            setProps
                [ ("font", pToJSVal $ fontString style font)
                , ("text", pToJSVal txt)
                ]
                obj
            retVal obj
```

... and then call that function within the `Text` and `StyledText` case alternatives, like this:

```haskell
    Text cs txt -> renderStyledText cs Plain Serif txt
    StyledText cs style font txt -> renderStyledText cs style font txt
```

kammitama5:
Ohhhh....I just saw this! Thank you so much!

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -574,10 +575,18 @@ picToObj' pic =
...
-                [ ("font", pToJSVal $ fontString style font)
+                [ ("font", pToJSVal $ fontString Plain Serif)
+                , ("text", pToJSVal txt)
+                ]
+                obj
+            retVal obj
+        StyledText cs style font txt -> do
+            obj <- init "styledText"
+            setProps
+                [ ("font", pToJSVal $ fontString Plain Serif)
```
cdsmith: I don't think you meant to say that all styledText is Plain and Serif.

kammitama5: Not a fan of Plain Serif? :D Made the changes. Thank you!

PR #641, opened 2018-05-23
Fixes #639: Use Node 8.x and run in Bash strict mode

### LINKED ISSUE #639: Installation on Ubuntu 18.04 fails due to Node 7 dependency
I'm still investigating the root cause, but I think [these](https://github.com/google/codeworld/blob/master/install.sh#L83) lines from `install.sh` are the problem. The requested version of Node cannot be installed on this version of Ubuntu. Unfortunately, the script continues regardless indicating two issues:

1. `install.sh` should fail fast if a dependency cannot be met
2. We should figure out how to get Node 7 on Ubuntu 18.04 or upgrade to a newer Node

rcook:
This change updates `install.sh` to install Node 8 instead of (now-deprecated) Node 7. It also runs using `set -euo pipefail` ("strict mode") in order to fail fast. I also added a `--quiet` switch to the `run` function to get rid of errors when using `run` in a pipeline.

### DIFF
--- a/base.sh
+++ b/base.sh
@@ -20,10 +20,23 @@ export LANG=${LANG:-C.UTF-8}
 export PREFIX=$BUILD
 
 function run {
-  OLD_PWD=$PWD
-  cd $1
+  local old_pwd=$PWD
+  local temp_pwd
+  local quiet=0
+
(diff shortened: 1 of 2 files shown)


cdsmith:
Thanks for the fix.

PR #642, opened 2018-05-25
Blank Constructor

kammitama5:


codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -691,6 +693,7 @@ getPictureCS (Sector cs _ _ _) = cs
 getPictureCS (Arc cs _ _ _) = cs
 getPictureCS (ThickArc cs _ _ _ _) = cs
 getPictureCS (Text cs _) = cs
+getPictureCS (Blank _) = emptyCallStack
```
cdsmith: You have a call stack here.  I'm not sure why the original code doesn't have a CallStack for `Pictures`, but I think you should go ahead and produce the one you've got.

codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -108,6 +108,8 @@ data Picture
     | CoordinatePlane CallStack
     | Logo CallStack
     | Pictures [Picture]
+    | Blank Picture
```
cdsmith: Update `blank` below to  this?

    blank :: HasCallStack => Picture
    blank = Blank callStack

codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -108,6 +108,8 @@ data Picture
     | CoordinatePlane CallStack
     | Logo CallStack
     | Pictures [Picture]
+    | Blank CallStack
+
```
cdsmith: No need for a second blank line here.

kammitama5: Thanks!

codeworld-api/src/CodeWorld/Driver.hs
cdsmith: Were you going to return the stored call stack here?

kammitama5: sorry. I had initially compared it to Pictures, so was using emptyCallStack to get it to type check :/

PR #644, opened 2018-05-27
Rectangle Constructor

kammitama5:
this seems to make sense for Picture but can't get Driver to type check.
I tried a couple times and couldn't get Driver to work so I did a couple times of resetting and starting again.
Am I in the right direction?


Gabriel439:
@kammitama5: The way I would approach this is to define a `rectangleToPolygon` utility like this:

```haskell
rectangleToPolygon :: CallStack -> Double -> Double -> Picture
rectangleToPolygon cs w h = Polygon cs [(-w / 2, -h / 2), (w / 2, -h / 2), (w / 2, h / 2), (-w / 2, h / 2)]
```

... and then if there is some downstream function that pattern matches on `Rectangle`, you call it again to pattern match on the equivalent `Polygon`, like this:

```haskell
example (Rectangle cs w h) = example (rectangleToPolygon cs w h)
```

kammitama5:
Chris had mentioned something about a rectangle being a rectangle and not a polygon.

codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -197,24 +210,23 @@ solidLoop :: HasCallStack => [Point] -> Picture
...
 
 {-# WARNING solidLoop "Please use solidClosedCurve instead of solidLoop." #-}
+rectangleToPolygon :: CallStack -> Double -> Double -> Picture
+rectangleToPolygon cs w h = Polygon cs [(-w / 2, -h / 2), (w / 2, -h / 2), (w / 2, h / 2), (-w / 2, h / 2)]
 
 -- | A thin rectangle, with this width and height
 rectangle :: HasCallStack => Double -> Double -> Picture
-rectangle w h =
-    polygon [(-w / 2, -h / 2), (w / 2, -h / 2), (w / 2, h / 2), (-w / 2, h / 2)]
+rectangle w h =
```
Gabriel439: All of these user-facing utilities like `rectangle`/`solidRectangle`/`thickRectangle` should follow the exact same pattern, which is:

```haskell
lowerCaseName = UpperCaseName callStack
```

In other words:

```haskell
rectangle :: CallStack -> Double -> Double -> Picture
rectangle = Rectangle callStack

solidRectangle :: HasCallStack => Double -> Double -> Picture
solidRectangle = SolidRectangle callStack

thickRectangle :: HasCallStack => Double -> Double -> Double -> Picture
thickRectangle = ThickRectangle callStack
```

Just stick to that pattern.  If you think you need to do something different, that's a good time to ask for help

kammitama5: Aye aye Cap'n. I'll work on it tomorrow :D

kammitama5: so I think I'm not sure about how to use the helper function and make a new constructor. I can get two of the functions to type check, but I'm still using the helper function, which uses Polygon :/

codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -57,6 +57,15 @@ data Picture
     | ThickPolygon CallStack
            [Point]
            !Double
+    | Rectangle CallStack
```
cdsmith: Let's start with getting these constructors right.  The arguments to each constructor in `Picture` should be a `CallStack`, and then the same arguments as the API function.  So `Rectangle` and `SolidRectangle` shouldn't have a list of points, just a width and height.  `ThickRectangle` needs a width, height, and line thickness.  Once you have that information, that's already enough to find the points!

I think you had this right at first, but instead of pushing forward and fixing the rest of the code, you keep going back and changing what you've already done.  That makes it hard to figure out where you're confused, and what you know.  So let's start here; get these types right, and the rest should follow from that.

codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -197,24 +206,21 @@ solidLoop :: HasCallStack => [Point] -> Picture
 solidLoop ps = SolidClosedCurve callStack ps
 
 {-# WARNING solidLoop "Please use solidClosedCurve instead of solidLoop." #-}
+rectangleToPolygon :: CallStack -> Double -> Double -> Picture
```
cdsmith: Did you mean to group this together with the WARNING pragma?  Why?

cdsmith: I'll disagree slightly with Gabe's comment.  The helper function I think you want here is this:

    rectangleVertices :: Double -> Double -> [Point]
    rectangleVertices w h = [ (w / 2, h / 2), (w / 2, -h / 2), (-w / 2, -h / 2), (-w / 2, h / 2) ]

I believe you want this in CodeWorld.Driver.  Then you can do something like this:

    pictureToDrawing (Rectangle _ w h) = Shape $ pathDrawer (rectangleVertices w h) 0 True False

and something similar for the other three constructors.

kammitama5: didn't mean to. Removed it.

codeworld-api/src/CodeWorld/Picture.hs
kammitama5: I think I'm a lot closer now, but just one more error to fix regarding points not being in scope in Driver.hs (lines 489, 500 and 505) (from last commit)

codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -213,16 +213,15 @@ rectangleVertices w h = [ (w / 2, h / 2), (w / 2, -h / 2), (-w / 2, -h / 2), (-w
 
 -- | A thin rectangle, with this width and height
 rectangle :: HasCallStack => Double -> Double -> Picture
-rectangle ps = rectangleToPolygon callStack ps
+rectangle w h = Rectangle callStack w h
```
cdsmith: This is correct if you want to leave it this way.  But just as a comment, notice that you can eta reduce these to:

    rectangle = Rectangle callStack

which is the pattern Gabe was suggesting you follow.  It means the same thing as what you wrote, but you're defining the whole function at once by partially applying the constructor, instead of defining it in terms of its arguments.

kammitama5: Ohhh..I see it now. Thank you!

kammitama5: so it seems like lines 487 to 508 should mention width and height somewhere...for example, Rectangle mentions only "width" on line 491. But it type-checked.

codeworld-api/src/CodeWorld/Picture.hs
kammitama5: I made a second version that has "width" and "linewidth".

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -481,6 +484,38 @@ picToObj' pic =
                 ]
                 obj
             retVal obj
+        Rectangle _ w h -> do
+            obj <- init "rectangle"
+            setProps
+                [ ("width",  pToJSVal (0 :: Double))
```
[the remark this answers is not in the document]
kammitama5: Okey doke. I'll fix it this evening. Thank you!! :D

cdsmith:
It looks like you're finished with this?  It looks good to me, so I'll merge it.  If you were still working on it, sorry!  (You can always open a new pull request for further changes.)

Issue #647, opened 2018-06-02
Code World produces error when ```/.run.sh```


kammitama5:
Currently using Ubuntu 18.04 on one laptop and 14.04 on the other. Both builds did not work.

Error message is:
```
Declaration for circle
Unfolding of circle:
  Can't find interface-file declaration for variable circle5
    Probable cause: bug in .hi-boot file, or inconsistent .hi file
    Use -ddump-if-trace to get an idea of which file caused the error
exported function not found: Fun {funPackage = Package {unPackage = "codeworld-api-0.2.2.1-KyKep1k46KT8IxuxPAKbD0"}, funModule = "CodeWorld.Picture", funSymbol = "h$codeworldzmapizm0zi2zi2zi1zmKyKep1k46KT8IxuxPAKbD0ZCCodeWorldziPicturezicircle1"}
When evaluating:
  error, called  in ghcjs-0.2.1-9zafNjYOdSgHPwMwF4xvpW:Gen2.Linker
```

![screenshot_2018-06-01_22-29-51](https://user-images.githubusercontent.com/8516378/40870022-e61a5142-65da-11e8-8836-cfb4f4f00e3b.png)

cdsmith:
Just guessing, but it looks like your package database might have a problem?  Note that I just changed the version number of `codeworld-api` from 0.2.2.1 to 0.2.3, so if you pulled from source, you might have some old state sitting around.  But it didn't happen in the live site, so I'm not sure what could be going on...

kammitama5:
So this works now... :dancer:

PR #648, opened 2018-06-02
Build failing code committed

kammitama5:
Hi Chris, these are the changes we made before we were testing ```./run.sh``` and had a buggy interface ie the interface loaded but when we tried to compile the program, we got the error in this [issue](https://github.com/google/codeworld/issues/647)


cdsmith:
This seems to be left over from old changes.  I'll close the pull request.

PR #658, opened 2018-06-11
printPrimitive print

kammitama5:
This is the starting point for trying to print a constructor primitive to UI. Assistance would be appreciated.


kammitama5:
screenshot from 2018-06-11 18-57-50

cdsmith:
Great!  This is exactly what I would have expected to see.  You're using the derived `Show` instance for `Picture`, which is going to print the value in Haskell syntax.  That's what you see: the `Circle` constructor, a call stack (which is most of the value), and then the radius, 2.0, at the end.

So the next task is to make that string say what you want.  I confess I'm not entirely sure *what* you want it to say, so a good step would be to try to describe clearly how you want this to work.

kammitama5:
I was thinking something to the extent of Circle (Double, Double) but maybe it's worth discussing at our next meeting. That's what I was thinking on my own, but it's worth musing about a solid consensus.

kammitama5:
If this commit is in the right direction, I can do the rest pretty quickly and then start making test functions to test the output of each.

kammitama5:
So I added another one and this is sort of what I have right now. But it's not really a drill-down.
But considering what shows up on the lhs text editor is the actual parameter, it might work?
screenshot from 2018-06-13 01-03-48

cdsmith:
I thought that you wanted to include the actual values.  So something like:

    import Text.Printf
    ...
    describePicture (Circle _ r) = printf "circle { radius = %.4f }" r

(I didn't test that, and haven't used printf in a while, and the types are kind of funky, so it might need a type annotation somewhere.)

kammitama5:
Oooo. I didn't think about that! Thanks so much! I'll definitely do that today!

kammitama5:
Okey doke. How do I do multiple parameters? I Thought it was something like ```"rectangle { width = %.4f} {height = %.4f}" w h``` but that's not working :/

cdsmith:
What do you mean by "not working"?  What's the error message?

kammitama5:
Would this work? Is there any reason why this might end up being too lengthy (say, if we have 10 arguments to the function)?

screenshot from 2018-06-13 19-36-00

kammitama5:
screenshot from 2018-06-13 20-04-56

kammitama5:
For arc angle precision...4 digits?
arc

cdsmith:
We're going to run into a problem here.  As you might recall, there are two CodeWorld modes: http://code.world, and http://code.world/haskell.  The form of these shapes is always a little different between the two, but with arcs, the difference is too great to ignore: arcs in the first are given in degrees; but in the second, they are given in radians.  It won't work for my students to show all angles in radians; at 11, they won't have any idea what they are looking at.  It's also probably not acceptable to show all angles in degrees when users of http://code.world/haskell are accustomed to working in radians.

This implies that what you show here should depend on the **mode**.  I can think of two ways to do this, but neither is trivial.

I'd suggest you not worry about this yet, and get your first prototype working with whatever is easiest.  Are we meeting tonight?  If so, we can talk then about what the right answer is.

kammitama5:
Yeah..that's a pretty big difference :/ I was wondering about that also, and had gotten just visible points the first time I used radians (because that one wanted degrees). Okay, I'll work around it for now and finish up polyline, etc in the meantime.

kammitama5:
I keep getting ```Polyline``` to resolve to ```Text``` using the ```intercalate``` function and I need it to resolve to ```String```. How can I do this? (see last commit)

cdsmith:
Are you using the `intercalate` from Data.List?  You'll have to add it to the import statement at the top.

kammitama5:
Oh, it works now. I had to be more specific about *which* intercalate :/

kammitama5:
So...how do I resolve these with the picture type? (ie ```rotate, colour, scale```)? Thanks! :D

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -49,13 +49,13 @@ import Control.Exception
...
 import Data.Char (chr)
-import Data.List (find, zip4)
+import Data.List (find, zip4, intercalate)
 import Data.Maybe (fromMaybe, isNothing, mapMaybe)
 import Data.Monoid
 import Data.Serialize
 import Data.Serialize.Text
 import qualified Data.Text as T
-import Data.Text (Text, pack, singleton)
+import Data.Text (Text, pack, singleton, intercalate)
```
cdsmith: I suggest not importing `Data.Text.intercalate`.  Then you can use `intercalate` unqualified on lists.  (If you DO need `Data.Text.intercalate`, you can already use it as `T.intercalate` because of the import on line 57)

kammitama5:
I'm also having issues with "StyledText".I keep getting 
```
• Couldn't match expected type ‘TextStyle
                                    -> Font -> Text -> String’
                  with actual type ‘[Char]’
    • The function ‘show’ is applied to four arguments,
      but its type ‘[Char] -> [Char]’ has only one

```
....but StyledText has three arguments...so....I'll push up that code. (currently commented out)

codeworld-api/src/CodeWorld/Picture.hs
```diff
@@ -127,20 +127,20 @@ data Picture
     | CoordinatePlane CallStack
     | Logo CallStack
     | Pictures [Picture]
-    | Blank CallStack
+    | Blank CallStack deriving (Show)
```
cdsmith: I don't think you need this `Show` instance any more.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -728,11 +729,44 @@ picToObj' pic =
...
+describePicture (ThickPolyline _ pts w) = printf "thickPolyline" ++ Data.List.intercalate "," [printf "(%4f, %4f)" x y | (x,y) <- pts] ++ printf ", {width = %4f}" w
+describePicture (Curve _ pts) = printf "thickCurve" ++ Data.List.intercalate "," [printf "(%4f, %4f)" x y | (x,y) <- pts]
+describePicture (ThickCurve _ pts w) = printf "thickCurve " ++ Data.List.intercalate "," [printf "(%4f, %4f)" x y | (x,y) <- pts] ++ printf ", {width = %4f}" w
+describePicture (Sector _ b e r) = printf "sector { startAngle = %4f , endAngle = %4f , radius = %4f}" b e r
+describePicture (Arc _ b e r) = printf "arc { startAngle = %4f , endAngle = %4f , radius = %4f}" b e r
+describePicture (ThickArc _ b e r w) = printf "thickArc { startAngle = %4f , endAngle = %4f , radius = %4f , width = %4f}" b e r w
+describePicture (Text _ txt) = printf "text { text = %s }" txt
+describePicture (Blank _) = printf "blank "
+--describePicture (StyledText _ style font txt) = style font txt
+--describePicture (Color cs _ _) = cs
```
cdsmith: For these transformations, you should probably just ignore the picture argument.  It will by underneath the transformation in the tree anyway.  So for Color, you'll want to just write something like: `"colored { color = RGBA(1, 1, 1, 1) }"`, and so on.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -728,11 +729,44 @@ picToObj' pic =
...
+--describePicture (Translate cs _ _ _) = cs
+--describePicture (Scale _ x y p) = show "scale { x = %s , y = %s , p = %s }" x y p
```
cdsmith: I split off `Dilated` as a separate constructor in a commit of my own.  You'll need a case for that, as well.

kammitama5:
Thank you! I'll get to work on this!

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -752,7 +752,7 @@ describePicture (Arc _ b e r) = printf "arc { startAngle = %4f , endAngle = %4f
 describePicture (ThickArc _ b e r w) = printf "thickArc { startAngle = %4f , endAngle = %4f , radius = %4f , width = %4f}" b e r w
 describePicture (Text _ txt) = printf "text { text = %s }" txt
 describePicture (Blank _) = printf "blank "
---describePicture (StyledText _ style font txt) = style font txt
+--describePicture (StyledText _ style font txt) = show " { style = %s , font = %s , txt = %s }" style font txt
```
cdsmith: The font and style are not strings, so %s won't work with them directly.  However, you have a `Show` instance, so this would work if you just change `font` to `(show font)`, and `style` to `(show style)`

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -728,11 +729,45 @@ picToObj' pic =
...
+describePicture (ThickPolyline _ pts w) = printf "thickPolyline" ++ intercalate "," [printf "(%4f, %4f)" x y | (x,y) <- pts] ++ printf ", {width = %4f}" w
+describePicture (Curve _ pts) = printf "thickCurve" ++ intercalate "," [printf "(%4f, %4f)" x y | (x,y) <- pts]
+describePicture (ThickCurve _ pts w) = printf "thickCurve " ++ intercalate "," [printf "(%4f, %4f)" x y | (x,y) <- pts] ++ printf ", {width = %4f}" w
+describePicture (Sector _ b e r) = printf "sector { startAngle = %4f , endAngle = %4f , radius = %4f}" b e r
+describePicture (Arc _ b e r) = printf "arc { startAngle = %4f , endAngle = %4f , radius = %4f}" b e r
+describePicture (ThickArc _ b e r w) = printf "thickArc { startAngle = %4f , endAngle = %4f , radius = %4f , width = %4f}" b e r w
+describePicture (Text _ txt) = printf "text { text = %s }" txt
+describePicture (Blank _) = printf "blank "
+describePicture (StyledText _ style font txt) = printf " { style = %s , font = %s , txt = %s }" (show style) (show font) txt
+--describePicture (Color _ (RGBA(r g b a)) _) = "colored { color = RGBA(1, 1, 1, 1) }" RGBA(r g b a)
```
cdsmith: I think you want something like this:

    describePicture (Color _ (RGBA r g b a) _) = printf "colored { color = RGBA(%4f, %4f, %4f, %4f) }" r g b a

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -728,11 +729,45 @@ picToObj' pic =
             return $ (unsafeCoerce arr :: JSVal)
     setProps xs obj = liftIO $ void $ mapM (\(s, v) -> setProp s v obj) xs
 
+-- describePicture 
+describePicture :: Picture -> String
+describePicture (Rectangle _ w h) = printf "rectangle { width = %4f , height = %4f }" w h
+describePicture (SolidPolygon _ pts) = printf "solidPolygon " ++ intercalate "," [printf "(%4f, %4f)" x y | (x,y) <- pts]
```
cdsmith: It would be great to stick to a consistent format, by adding braces and a name for the parameter.  Something like this:

    describePicture (SolidPolygon _ pts) = printf "solidPolygon { points = %s }" (intercalate ", " [printf "(%4f, %4f)" x y | (x,y) <- pts])

The same comment applies to other cases where you do something similar.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -728,11 +729,45 @@ picToObj' pic =
             return $ (unsafeCoerce arr :: JSVal)
     setProps xs obj = liftIO $ void $ mapM (\(s, v) -> setProp s v obj) xs
 
+-- describePicture
```
cdsmith: I don't think this comment adds anything.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -728,11 +729,45 @@ picToObj' pic =
...
+describePicture (Circle _ r) = printf "circle { radius = %4f }" r
+describePicture (SolidCircle _ r) = printf "solidCircle { radius = %4f }" r
+describePicture (ThickCircle _ lw r) = printf "thickCircle { linewidth = %4f , radius = %4f }" lw r
+describePicture (Polyline _ pts) = printf "polyline " ++ intercalate "," [printf "(%4f, %4f)" x y | (x,y) <- pts]
+describePicture (ThickPolyline _ pts w) = printf "thickPolyline" ++ intercalate "," [printf "(%4f, %4f)" x y | (x,y) <- pts] ++ printf ", {width = %4f}" w
+describePicture (Curve _ pts) = printf "thickCurve" ++ intercalate "," [printf "(%4f, %4f)" x y | (x,y) <- pts]
+describePicture (ThickCurve _ pts w) = printf "thickCurve " ++ intercalate "," [printf "(%4f, %4f)" x y | (x,y) <- pts] ++ printf ", {width = %4f}" w
+describePicture (Sector _ b e r) = printf "sector { startAngle = %4f , endAngle = %4f , radius = %4f}" b e r
+describePicture (Arc _ b e r) = printf "arc { startAngle = %4f , endAngle = %4f , radius = %4f}" b e r
+describePicture (ThickArc _ b e r w) = printf "thickArc { startAngle = %4f , endAngle = %4f , radius = %4f , width = %4f}" b e r w
```
cdsmith: Maybe call this `linewidth`, for consistency?

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -728,11 +729,45 @@ picToObj' pic =
...
+describePicture (Text _ txt) = printf "text { text = %s }" txt
```
cdsmith: I think it makes more sense to show the string in quotes.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -728,11 +729,45 @@ picToObj' pic =
...
+describePicture (Blank _) = printf "blank "
```
cdsmith: Why the trailing space?  Same comment for other similar cases.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -728,11 +729,45 @@ picToObj' pic =
...
+describePicture (StyledText _ style font txt) = printf " { style = %s , font = %s , txt = %s }" (show style) (show font) txt
+--describePicture (Color _ (RGBA(r g b a)) _) = "colored { color = RGBA(1, 1, 1, 1) }" RGBA(r g b a)
+describePicture (Translate _ x y _) = printf "translate { x = %4f , y = %4f }" x y
```
cdsmith: Let's use the public API name, which is `translated`.  Same for other transformations.

(The constructor names being different is a historical oddity, but no one ever sees these non-exported constructor names.)

codeworld-api/src/CodeWorld/Driver.hs
kammitama5: for ```thickCurve``` or ```thickArc``` or both?

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -243,6 +243,8 @@ pictureToDrawing (Scale _ x y p) =
     Transformation (scaleDS x y) $ pictureToDrawing p
 pictureToDrawing (Rotate _ r p) =
     Transformation (rotateDS r) $ pictureToDrawing p
+pictureToDrawing (Dilated _ k p) =
```
cdsmith: If you pull from github (`git pull --rebase`), then this should already be added.

kammitama5: thanks!

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -782,11 +783,43 @@ picToObj' pic =
...
+describePicture (ThickArc _ b e r w) = printf "thickArc { startAngle = %4f , endAngle = %4f , radius = %4f , width = %4f}" b e r w
+describePicture (Text _ txt) = printf "text { text = '%s' }" txt
+describePicture (Blank _) = printf "blank"
+describePicture (StyledText _ style font txt) = printf " styledText { style = %s , font = %s , txt = '%s' }" (show style) (show font) txt
+describePicture (Color _ (RGBA r g b a) _) = printf "colored { color = RGBA(%4f, %4f, %4f, %4f) }" r g b a
+describePicture (Translate _ x y _) = printf "translated { x = %4f , y = %4f }" x y
+describePicture (Scale _ x y _) = printf "scaled { x = %4f , y = %4f }" x y
+describePicture (Rotate _ angle _) = printf "rotated { angle = %4f }" angle
+describePicture (Dilate _ k _) = printf "dilated { number = %4f }" k
+describePicture (Logo _) = printf "logo"
```
cdsmith: The public API here is `codeWorldLogo`

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -782,11 +783,43 @@ picToObj' pic =
...
+describePicture (Arc _ b e r) = printf "arc { startAngle = %4f , endAngle = %4f , radius = %4f}" b e r
+describePicture (ThickArc _ b e r w) = printf "thickArc { startAngle = %4f , endAngle = %4f , radius = %4f , width = %4f}" b e r w
+describePicture (Text _ txt) = printf "text { text = '%s' }" txt
+describePicture (Blank _) = printf "blank"
+describePicture (StyledText _ style font txt) = printf " styledText { style = %s , font = %s , txt = '%s' }" (show style) (show font) txt
+describePicture (Color _ (RGBA r g b a) _) = printf "colored { color = RGBA(%4f, %4f, %4f, %4f) }" r g b a
+describePicture (Translate _ x y _) = printf "translated { x = %4f , y = %4f }" x y
+describePicture (Scale _ x y _) = printf "scaled { x = %4f , y = %4f }" x y
+describePicture (Rotate _ angle _) = printf "rotated { angle = %4f }" angle
+describePicture (Dilate _ k _) = printf "dilated { number = %4f }" k
```
cdsmith: Maybe "factor" instead of "number"?

kammitama5:
Hmm..so how should I go about with multiple vertices? Any ideas?

Gabriel439:
@kammitama5: I would suggest the following really general solution:

Write a function called `trim` with this type:

```haskell
trim :: Int -> String -> String
```

The first argument to `trim` is an `Integer` specifying a maximum display length for the `String`.  The second argument to `trim` is the input `String` that we want to trim to fit within the specified maximum display length.

If the input `String` is already less than the specified display length then `trim` does nothing and returns the input `String` unmodified as the output `String`.  If the input `String` is longer than the specified display length then trim replaces the interior of the string using `...`.  The number of characters to replace in the interior is chosen such that the trimmed string fits within the specified maximum display length.

Once you have that `trim` function, then you can use it to transform the result of `describePicture` to not exceed some specified length.

The other reason to do things this way is that you could then later add another change that lets the user switch between the trimmed and untrimmed output.

kammitama5:
Thanks!

kammitama5:
the naming could be more meaningful, but....comments?

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -782,11 +783,53 @@ picToObj' pic =
...
 
+-- x is the maximum display length (Int)
+-- y is the input string we want to trim
+-- if string (y) is less than x, then we return y
+-- otherwise, replace the interior of string with ...
+
+trim :: Int -> String -> String
+trim x y = case (x > (length y)) of
+      True -> y
+      False -> "..."
```
Gabriel439: For the case where `y` is too large, you probably still want to render part of the string.  The idea is that you could just chop out the middle, like this:

```
Vertices [1.0, 2.0, 3.0, 7...9.0, 1.0]
```

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -782,11 +783,53 @@ picToObj' pic =
             return $ (unsafeCoerce arr :: JSVal)
     setProps xs obj = liftIO $ void $ mapM (\(s, v) -> setProp s v obj) xs
 
+-- x is the maximum display length (Int)
+-- y is the input string we want to trim
+-- if string (y) is less than x, then we return y
+-- otherwise, replace the interior of string with ...
+
+trim :: Int -> String -> String
+trim x y = case (x > (length y)) of
```
Gabriel439: Also, `case (x > (length y)) of True -> ...; False` is the same as an `if` expression:

```haskell
trim x y = if x > length y then y else ...
```

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -782,11 +783,53 @@ picToObj' pic =
...
+describePicture (CoordinatePlane _) = printf "coordinatePlane"
+describePicture (Pictures _) = printf "pictures"
+
 setCallInfo :: Picture -> Object -> IO ()
 setCallInfo pic obj =
     case findCSMain (getPictureCS pic) of
-        Just (callName, src) -> do
-            setProp "name" (pToJSVal $ callName) obj
+        Just (_, src) -> do
+            setProp "name" (pToJSVal $ describePicture pic) obj
```
Gabriel439: Also, don't forget to use the `trim` function down here

codeworld-api/src/CodeWorld/Driver.hs
kammitama5: Oh, so basically take from the first half of the string (that is longer than allotted string length), and the second half, accounting for the "..." length.

Gabriel439: @kammitama5: Yeah

codeworld-api/src/CodeWorld/Driver.hs
kammitama5: So I think the function works but I may be missing some edge cases. Can you give me a teeny tiny hint as to where it should go? :D

kammitama5:
this is without the truncation (say, string length is less than x)
screenshot from 2018-06-19 05-01-42

this is with the truncation
screenshot from 2018-06-19 05-00-36

I set an arbitrary number of 30 (for max string length) for now, but that can be changed.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -782,11 +783,49 @@ picToObj' pic =
             return $ (unsafeCoerce arr :: JSVal)
     setProps xs obj = liftIO $ void $ mapM (\(s, v) -> setProp s v obj) xs
 
+trim :: Int -> String -> String
```
cdsmith: Nitpicking here, but:

* You don't need the outer parentheses when defining `mid`
* You don't need parentheses around the expression between `case` and `of`
* You don't need the outer parentheses on either branch of the `case` (lines 789 and 790).

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -782,11 +783,49 @@ picToObj' pic =
...
+describePicture(ThickClosedCurve _ pts w) = printf "thickClosedCurve { points = %s, width = %4f }" (intercalate ", " [printf "(%4f, %4f)" x y | (x,y) <- pts]) w
+describePicture (Circle _ r) = printf "circle { radius = %4f }" r
+describePicture (SolidCircle _ r) = printf "solidCircle { radius = %4f }" r
+describePicture (ThickCircle _ lw r) = printf "thickCircle { linewidth = %4f , radius = %4f }" lw r
+describePicture (Polyline _ pts) = printf "polyline { points = %s }" (intercalate ", " [printf "(%4f, %4f)" x y | (x,y) <- pts])
+describePicture (ThickPolyline _ pts w) = printf "thickPolyline { points = %s, width = %4f }" (intercalate ", " [printf "(%4f, %4f)" x y | (x,y) <- pts]) w
+describePicture (Curve _ pts) = printf "curve { points = %s }" (intercalate ", " [printf "(%4f, %4f)" x y | (x,y) <- pts])
+describePicture (ThickCurve _ pts w) = printf "thickCurve { points = %s, width = %4f }" (intercalate ", " [printf "(%4f, %4f)" x y | (x,y) <- pts]) w
+describePicture (Sector _ b e r) = printf "sector { startAngle = %4f , endAngle = %4f , radius = %4f}" b e r
```
cdsmith: I'm still not sure what to do about angles.  Plumbing the mode through seems painful.  Perhaps you can show them in both degrees and radians?  Like this:

    startAngle = 30.0° (0.52 radians)

(You can compute degrees as `180 * angle / pi`)

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -782,11 +783,49 @@ picToObj' pic =
...
+describePicture (CoordinatePlane _) = printf "coordinatePlane"
+describePicture (Pictures _) = printf "pictures"
+
 setCallInfo :: Picture -> Object -> IO ()
 setCallInfo pic obj =
     case findCSMain (getPictureCS pic) of
-        Just (callName, src) -> do
-            setProp "name" (pToJSVal $ callName) obj
+        Just (_, src) -> do
+            setProp "name" (pToJSVal $ (trim 30 . describePicture) pic) obj
```
cdsmith: I'd pick a larger number than 30 here.  Maybe 80?

kammitama5: Sure. Sounds good.

kammitama5:
screenshot from 2018-06-19 13-54-48
Yep, it looks better with 80.

kammitama5:
I was getting a really long trail. I used "%.2g". Would you prefer trail or the abridged one (ie to 2 dp) for angles?
screenshot from 2018-06-19 14-10-04
screenshot from 2018-06-19 14-48-16

kammitama5:
so the degrees and radians are to two dp, but the other parts are not. Is this ok?

PR #671, opened 2018-06-21
ShowShortFloat function

kammitama5:
ShowShortFloat function


codeworld-base/src/Internal/Num.hs
```diff
@@ -128,6 +128,12 @@ instance P.Show Number where
         stripZeros =
             P.reverse . P.dropWhile (== '.') . P.dropWhile (== '0') . P.reverse
 
+instance P.ShowShortFloat x where
```
cdsmith: The idea was to copy the `show` function from this instance, but not to define a new instance of your own.  An instance would need to be an instance of some type class, and there's no class called `ShowShortFloat` to have an instance of.

I think you just want a plain function, like this:

    showShortFloat :: Double -> String
    showShortFloat x = stripZeros (showFFloatAlt (Just 4) x "")
      where stripZeros = reverse . dropWhile (== '.') . dropWhile (== '0') . reverse

You want this in CodeWorld.Driver, not here.  This file doesn't need to change.

kammitama5:
So currently my instance ```ShowShortFloat``` is out of scope because it's in the Codeworld/Codeworld-base/Internal/Number.hs file. I'm currently in ```Driver.hs``` trying to go up/down and into the file structure to get it into scope from ```Codeworld-base``` to ```codeworld-api/Driver.hs```. How do I do this using ```import qualified as```? Thank you!

codeworld-base/src/Internal/Num.hs
kammitama5: Thanks. I misunderstood. It would still be nice to know how to go up and down the file structure also using import.

cdsmith: I'm not sure I understand what you're asking.  There are two different packages here: `codeworld-api` and `codeworld-base`.  You can't import stuff from `codeworld-base` into `codeworld-api`, because codeworld-api doesn't depend on codeworld-base.  The dependencies of codeworld-api are listed in https://github.com/google/codeworld/blob/master/codeworld-api/codeworld-api.cabal  But you also can't make `codeworld-api` depend on `codeworld-base`, because `codeworld-base` depends on `codeworld-api` already.  If you added the other direction, you'd have a dependency cycles, and neither package could be built because the other one needs to be built first.

This is precisely why I was suggesting that you copy the code here over to where you need it, instead of trying to reuse one implementation in both places.  There simply isn't a good place yet to put common utility code that's shared by both `codeworld-api` and `codeworld-base`.

kammitama5: I see. Makes sense. Thanks!

kammitama5:
rectangle function test 

screenshot from 2018-06-21 07-54-21

kammitama5:
So I have a question. I've been trying to figure out the ```intercalate``` with two arguments. I understand with one argument ```intercalate "," [ ]``` but I'm unsure how to use ```showShortFloat``` with the list comprehension.

cdsmith:
I'm not sure how to answer, but I'll try.  `intercalate` always takes two arguments: the separator (","), and the list.  I don't know what you mean by one argument `intercalate`.

The way you'd use `showShortFloat` is by changing the list comprehension, as in: `intercalate ", " [ "(" ++ shorShortFloat x ++ ", " ++ showShortFloat y ++ ")" | (x, y) <- points ]`

kammitama5:
that's really helpful! Thank you. I should have been more specific. I meant that ```showShortFloat``` takes one argument, but I wanted to apply it to every argument of the list-comprehension (ie x and y) in the intercalate function while keeping the formatting.

kammitama5:
Any recommendations for the remaining functions? ie ```Text```, ```StyledText```, ```Blank```, ```Logo```, etc? Will those stay the same or should I remove the ```printf```?

cdsmith:
I think you should remove the printf where you don't need it, such as in `codeWorldLogo`.  It doesn't do anything there, since there are no patterns to substitute.  Leaving `printf` in the other cases where you use it is fine.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -789,35 +789,38 @@ trim x y = let mid = (x - 2) `div` 2
...
+describePicture (Polygon _ pts) = "polygon { points = " ++ intercalate ", " [ "(" ++ showShortFloat x ++ ", " ++ showShortFloat y ++ ")" | (x, y) <- pts ] ++ " }"
+describePicture (ThickPolygon _ pts w) = "thickPolygon { points = " ++ intercalate ", " [ "(" ++ showShortFloat x ++ ", " ++ showShortFloat y ++ ")" | (x, y) <- pts ] ++ ", width = " ++ showShortFloat w ++ " }"
+describePicture (SolidRectangle _ w h) = "solidRectangle { width = " ++ showShortFloat w ++ " , height = " ++ showShortFloat h ++ "}"
+describePicture (ThickRectangle _ lw w h) = "thickRectangle { linewidth = " ++ showShortFloat lw ++ " , width = " ++ showShortFloat w ++ ", height = " ++ showShortFloat h ++ " }"
+describePicture (ClosedCurve _ pts) = "closedCurve { points = " ++ intercalate ", " [ "(" ++ showShortFloat x ++ ", " ++ showShortFloat y ++ ")" | (x, y) <- pts ] ++ " }"
+describePicture(ThickClosedCurve _ pts w) = "thickClosedCurve { points = " ++ intercalate ", " [ "(" ++ showShortFloat x ++ ", " ++ showShortFloat y ++ ")" | (x, y) <- pts ] ++ ", width = " ++ showShortFloat w ++ " }"
+describePicture (Circle _ r) = "circle { radius = " ++ showShortFloat r ++ "}"
```
cdsmith: Here and in the next few: everywhere else, we leave a space before the closing brace.  Let's do the same here.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -789,35 +789,38 @@ trim x y = let mid = (x - 2) `div` 2
...
+describePicture (Sector _ b e r) =  "sector { startAngle = " ++ showShortFloat (180 * b / pi) ++ "° (" ++ showShortFloat b ++ " radians)" ++ " , endAngle = " ++ showShortFloat (180 * e / pi) ++ "°" ++ " (" ++ showShortFloat e ++ " radians ) ," ++ " radius = " ++ showShortFloat r ++ " }"
+describePicture (Arc _ b e r) = "arc { startAngle = " ++ showShortFloat (180 * b / pi) ++ "° (" ++ showShortFloat b ++ " radians)" ++ " , endAngle = " ++ showShortFloat (180 * e / pi) ++ "°" ++ " (" ++ showShortFloat e ++ " radians ) ," ++ " radius = " ++ showShortFloat r ++ " }"
+describePicture (ThickArc _ b e r w) =  "thickArc { startAngle = " ++ showShortFloat (180 * b / pi) ++ "° (" ++ showShortFloat b ++ " radians)" ++ " , endAngle = " ++ showShortFloat (180 * e / pi) ++ "°" ++ " (" ++ showShortFloat e ++ " radians ) ," ++ " radius = " ++ showShortFloat r ++ " , width = " ++ showShortFloat w ++ " }"
 describePicture (Text _ txt) = printf "text { text = '%s' }" txt
 describePicture (Blank _) = printf "blank"
```
cdsmith: printf isn't needed

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -789,35 +789,38 @@ trim x y = let mid = (x - 2) `div` 2
...
 describePicture (StyledText _ style font txt) = printf " styledText { style = %s , font = %s , txt = '%s' }" (show style) (show font) txt
-describePicture (Color _ (RGBA r g b a) _) = printf "colored { color = RGBA(%4f, %4f, %4f, %4f) }" r g b a
-describePicture (Translate _ x y _) = printf "translated { x = %4f , y = %4f }" x y
-describePicture (Scale _ x y _) = printf "scaled { x = %4f , y = %4f }" x y
-describePicture (Rotate _ angle _) = printf "rotated { angle = %4f }" angle
-describePicture (Dilate _ k _) = printf "dilated { factor = %4f }" k
-describePicture (Logo _) = printf "codeWorldLogo"
+describePicture (Color _ (RGBA r g b a) _) = "colored { color = RGBA( " ++ showShortFloat r ++ showShortFloat g ++ showShortFloat b ++ showShortFloat a ++ ") }"
+describePicture (Scale _ x y _) = "scaled { x = " ++ showShortFloat x ++ " , y = " ++ showShortFloat y ++ "}"
+describePicture (Rotate _ angle _) = "rotated { " ++ showShortFloat angle ++ "}"
```
cdsmith: Shouldn't there be "angle = " here?

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -789,35 +789,38 @@ trim x y = let mid = (x - 2) `div` 2
...
+describePicture (Dilate _ k _) = "dilated { factor = " ++ showShortFloat k ++  "}"
+describePicture (Logo _) = "codeWorldLogo"
 describePicture (CoordinatePlane _) = printf "coordinatePlane"
```
cdsmith: No need for printf

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -789,35 +789,38 @@ trim x y = let mid = (x - 2) `div` 2
...
 describePicture (Pictures _) = printf "pictures"
```
cdsmith: No need for printf

codeworld-api/src/CodeWorld/Driver.hs
kammitama5: I left it for text. Should I remove it? It doesn't type check (of type Text) if I just remote the ```printf```

kammitama5:
so the only two I have with printf are Text and Styled Text. Should I resolve this?

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -789,37 +789,40 @@ trim x y = let mid = (x - 2) `div` 2
...
+describePicture (Arc _ b e r) = "arc { startAngle = " ++ showShortFloat (180 * b / pi) ++ "° (" ++ showShortFloat b ++ " radians)" ++ ", endAngle = " ++ showShortFloat (180 * e / pi) ++ "°" ++ " (" ++ showShortFloat e ++ " radians )," ++ " radius = " ++ showShortFloat r ++ " }"
+describePicture (ThickArc _ b e r w) =  "thickArc { startAngle = " ++ showShortFloat (180 * b / pi) ++ "° (" ++ showShortFloat b ++ " radians)" ++ ", endAngle = " ++ showShortFloat (180 * e / pi) ++ "°" ++ " (" ++ showShortFloat e ++ " radians )," ++ " radius = " ++ showShortFloat r ++ ", width = " ++ showShortFloat w ++ " }"
 describePicture (Text _ txt) = printf "text { text = '%s' }" txt
-describePicture (Blank _) = printf "blank"
-describePicture (StyledText _ style font txt) = printf " styledText { style = %s , font = %s , txt = '%s' }" (show style) (show font) txt
-describePicture (Color _ (RGBA r g b a) _) = printf "colored { color = RGBA(%4f, %4f, %4f, %4f) }" r g b a
-describePicture (Translate _ x y _) = printf "translated { x = %4f , y = %4f }" x y
```
cdsmith: We lost the Translate constructor here somewhere.  I am adding it back in a different commit.

kammitama5: okey doke. Thank you.

PR #682, opened 2018-07-01
UI improvements to new SpeedSlider control

kammitama5:
this is a start. 
I am getting this error:
```
• No instance for (RealFloat a)
        arising from a use of ‘showFFloatAlt’
```

### DIFF
--- a/codeworld-api/src/CodeWorld/Driver.hs
+++ b/codeworld-api/src/CodeWorld/Driver.hs
@@ -2352,7 +2352,7 @@ drawControl w alpha SpeedSlider = translated (-4) (-9) p
     p =
         colored
             (RGBA 0 0 0 alpha)
-            (scaled 0.5 0.5 $ translated 0 0.5 $ lettering (pack (showFFloatAlt (Just 2) (playbackSpeed w) ""))) <>
+            (scaled 0.5 0.5 $ translated 0 0.8 $ lettering (pack (showFFloatAlt (Just 2) (playbackSpeed w) ""))) <>
         translated x 0 (solidRectangle 0.2 0.8) <>
         colored (RGBA 0.2 0.2 0.2 alpha) (rectangle 2.8 0.6) <>
         colored (RGBA 0.8 0.8 0.8 alpha) (solidRectangle 2.8 0.6)
--- a/codeworld-api/src/CodeWorld/Driver.hs
+++ b/codeworld-api/src/CodeWorld/Driver.hs
@@ -2352,7 +2352,7 @@ drawControl w alpha SpeedSlider = translated (-4) (-9) p
     p =
         colored
             (RGBA 0 0 0 alpha)
-            (scaled 0.5 0.5 $ lettering (pack (showFFloatAlt (Just 2) (playbackSpeed w) ""))) <>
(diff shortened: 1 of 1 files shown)


cdsmith:
That error happens because `state w` refers the state of the *underlying* simulation.  The `TimeLabel` control has type `Control Double` (see line 2242), so `state w` is guaranteed to have type `Double`.  But `SpeedSlider` has the type `Control a`, so you don't know what type `state w` will have.

This is okay, because you don't want to show the underlying state.  You want to show the playback speed instead!  Try changing `state w` to `playbackSpeed w`.

kammitama5:
Oh, I see. That's what it was referring to when it gave the error of ```RealFoat``` (ie Double). Thanks! Makes sense.

kammitama5:
Yes! There are numbers! :D :1st_place_medal:

kammitama5:
what's the next best step? Should I start the ```snap``` function?

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2258,25 +2259,28 @@ wrappedEvent ::
     -> Event
     -> Wrapped a
     -> Wrapped a
-wrappedEvent _ _ (MouseMovement _) w = w {mouseMovedTime = 0}
-wrappedEvent ctrls f (MousePress LeftButton p) w =
-    (foldr (handleControl f p) w (ctrls w)) {mouseMovedTime = 0}
-wrappedEvent _ _ _ w = w
+wrappedEvent _ _ (TimePassing _) w = w
+wrappedEvent ctrls f (event) w =
+    (foldr (handleControl f (event)) w (ctrls w)) {mouseMovedTime = 0}
```
cdsmith: There's no need for parentheses around `event`.  Parentheses around a single variable are always unneeded.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2231,6 +2231,7 @@ data Wrapped a = Wrapped
     { state :: a
     , playbackSpeed :: Double
     , mouseMovedTime :: Double
```
cdsmith: Consider renaming `mouseMovedTime` to something like `lastInteractionTime`?  It's now tracking time since any user event, not just mouse movement.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2258,25 +2259,28 @@ wrappedEvent ::
...
     | -8.4 < x && x < -7.6 && -9.4 < y && y < -8.6 = w {playbackSpeed = 0}
-handleControl _ (x, y) BackButton w
+handleControl _ (PointerPress (x,y)) BackButton w
     | -7.4 < x && x < -6.6 && -9.4 < y && y < -8.6 =
         w {state = max 0 (state w - 0.1)}
-handleControl f (x, y) StepButton w
+handleControl f (PointerPress (x, y)) StepButton w
     | -6.4 < x && x < -5.6 && -9.4 < y && y < -8.6 = w {state = f 0.1 (state w)}
-handleControl f (x, y) SpeedSlider w
+handleControl _ (PointerPress (x, y)) SpeedSlider w
```
cdsmith: It looks like you still haven't worked through the dragging logic here?

cdsmith:
I think the next step is to make dragging work as expected.

kammitama5:
thanks. I'll work on this tomorrow.

PR #688, opened 2018-07-10
Add debugSimulationOf with time-travelling feature

kammitama5:


cdsmith:
Cool.  I'll leave this pull request open, though, and we can merge once there's new functionality there.

kammitama5:
so this doesn't type-check, but I'm trying to work through the logic. ```debugSimulationOf``` takes an initial state and the step function, which generates a list of states, which are turned into pictures. Or is this the wrong path? Thanks!

cdsmith:
Oh, yes I think that's the wrong direction.  The *type* of `debugSimulationOf` tells you how a user (that is, a student) can use it.  We want a student to be able to replace `simulationOf` with `debugSimulationOf` without making any more changes, so the *type* needs to stay the same.

What you want to do instead is change the definition.

kammitama5:
Do you want me to put ```debugInteractionOf``` on a new branch and make a new PR?

cdsmith:
Let's keep working in this pull request for now.  So far, this new button isn't compelling enough to be worth increasing the API surface; so I'd rather not merge yet.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2126,7 +2111,7 @@ runPauseable initial stepHandler controls drawHandler = do
           where
             plainDrawing = pictureToDrawing $ drawHandler (state wrappedState)
         drawPicHandler (debugState, wrappedState) =
-            drawHandler $ state wrappedState
+            drawHandler $ state wrappedState
```
cdsmith: These extra spaces aren't needed.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2406,15 +2395,25 @@ wrappedStep f dt w =
...
+   | otherwise = fmap (eventHandler (TimePassing dt)) w
+wrappedEvent ctrls f eventHandler event w
+   | playbackSpeed w == 0 || handled = afterControls {lastInteractionTime = 0}
+   | otherwise = fmap (eventHandler event) afterControls {lastInteractionTime = 0}
+   where 
+         afterControls :: Wrapped a
+         handled :: Bool
+         (afterControls, handled) = foldr stepFunction (w, False) (ctrls w)
+         stepFunction :: Control a -> (Wrapped a, Bool) -> (Wrapped a, Bool)
+         stepFunction control (world, handled) = if handled then (world, True) else handleControl f event control world
```
cdsmith: wrap this line, please?  It's very long.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2540,12 +2552,41 @@ simulationControls w
     | playbackSpeed w == 0 = [PlayButton, StepButton, SpeedSlider]
     | otherwise = [PauseButton, SpeedSlider]
 
+debugSimulationControls :: Wrapped [w] -> [Control [w]]
```
cdsmith: Since these are used for both simulations and interactions, is there a better name?  Perhaps statefulDebugControls?  That's not great, but you get the idea...

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2540,12 +2552,41 @@ simulationControls w
...
   where
     initial =
         Wrapped {state = simInitial, playbackSpeed = 1, lastInteractionTime = 1000, isDragging = False}
 
+debugSimulationOf simInitial simStep simDraw =
+    runInspect debugSimulationControls initial step (\_ r -> r) draw
+  where
+    initial =
+        Wrapped {state = [simInitial], playbackSpeed = 1, lastInteractionTime = 1000, isDragging = False}
+    step dt (x:xs) = simStep dt x : x : xs
```
cdsmith: I wonder if you want to do the same thing here that you do in `debugInteractionOf`.  But it's not really important.  Simulations whose state doesn't change in `step` are very rare.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2540,12 +2552,41 @@ simulationControls w
...
+
 trace msg x = unsafePerformIO $ do
     hPutStrLn stderr (T.unpack msg)
     return x
+
+debugInteractionOf baseInitial baseStep baseEvent baseDraw = runInspect debugSimulationControls initial step event draw 
+  where
+    initial =
+        Wrapped {state = [baseInitial], playbackSpeed = 1, lastInteractionTime = 1000, isDragging = False}
+    event e (x:xs) = case x `seq` x' `seq` reallyUnsafePtrEquality# x x' of
```
cdsmith: Since this is non-trivial and repeated, it's probably worth a function.  Something like:

    prependIfChanged :: (a -> a) -> [a] -> [a]
    prependIfChanged f (x:xs) = case x `seq` x' `seq` ifTrue# (reallyUnsafePtrEquality# x x') of
        0# -> x' : x : xs
        _ -> x : xs
      where x' = f x

Then you have

    step dt = prependIfChanged (baseStep dt)
    event e = prependIfChanged (baseEvent e)

and you're done!

codeworld-api/src/CodeWorld/Driver.hs
kammitama5: I used ```statefulDebugControls```, but if there's a better name, can always change it.

codeworld-api/src/CodeWorld/Driver.hs
kammitama5: I changed it to be similar to ```debugInteractionOf``` but didn't add the ```prepend``` function here. I can do that also if needed.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2126,7 +2111,7 @@ runPauseable initial stepHandler controls drawHandler = do
           where
             plainDrawing = pictureToDrawing $ drawHandler (state wrappedState)
         drawPicHandler (debugState, wrappedState) =
-            drawHandler $ state wrappedState
+            drawHandler $ state wrappedState
```
cdsmith: Looks like you removed some of the trailing whitespace, but left some?

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2540,12 +2553,44 @@ simulationControls w
...
   where
     initial =
         Wrapped {state = simInitial, playbackSpeed = 1, lastInteractionTime = 1000, isDragging = False}
 
+debugSimulationOf simInitial simStep simDraw =
+    runInspect statefulDebugControls initial step (\_ r -> r) draw
+  where
+    initial =
+        Wrapped {state = [simInitial], playbackSpeed = 1, lastInteractionTime = 1000, isDragging = False}
+    step dt (x:xs) = case x `seq` x' `seq` reallyUnsafePtrEquality# x x' of
```
cdsmith: If you make prependIfChanged a top-level function, you can use it here too.

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2540,12 +2553,44 @@ simulationControls w
...
+        0# -> x' : x : xs
+        _  -> x : xs
+      where x' = simStep dt x
+    draw (x:xs) = simDraw x
+
 trace msg x = unsafePerformIO $ do
```
cdsmith: It doesn't make sense to put trace between the two debug entry points.  Put it back at the end of the file?

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2061,47 +2084,9 @@ runStatic pic = do
...
-    inspect (drawPicHandler <$> getState) pauseEvent highlightSelectEvent
-
-runPauseable ::
-       Wrapped s
-    -> (Double -> s -> s)
-    -> (Wrapped s -> [Control s])
-    -> (s -> Picture)
-    -> IO ()
-runPauseable initial stepHandler controls drawHandler = do
+runInspect ::
```
cdsmith: This type signature would make a lot more sense to me if the second parameter was an `s` rather than a `Wrapped s`.  You can do that by moving the initialization of Wrapped into this function.  I'd like to see that happen if you want to do it.  Otherwise, I'll do it myself after the merge.

(This isn't a new problem.  The type of the old `runPausable` was already pretty weird.)

codeworld-api/src/CodeWorld.hs
```diff
@@ -23,7 +23,9 @@ module CodeWorld (
     , groupActivityOf
     , unsafeGroupActivityOf
     , simulationOf
+    , debugSimulationOf
```
cdsmith: Do you want to add these to codeworld-base as well?

codeworld-api/src/CodeWorld/Driver.hs
```diff
@@ -2406,15 +2395,26 @@ wrappedStep f dt w =
...
+   | otherwise = fmap (eventHandler (TimePassing dt)) w
+wrappedEvent ctrls f eventHandler event w
+   | playbackSpeed w == 0 || handled = afterControls {lastInteractionTime = 0}
+   | otherwise = fmap (eventHandler event) afterControls {lastInteractionTime = 0}
+   where 
+         afterControls :: Wrapped a
+         handled :: Bool
+         (afterControls, handled) = foldr stepFunction (w, False) (ctrls w)
+         stepFunction :: Control a -> (Wrapped a, Bool) -> (Wrapped a, Bool)
+         stepFunction control (world, handled) =
```
cdsmith: I think stepFunction would be clearer with guards instead of an if.

codeworld-api/src/CodeWorld/Driver.hs
kammitama5: I'd like to understand how this works. I can try to work on this tomorrow and see if I can get it to work (ie before we meet).

cdsmith:
Okay, now I did the merge of my changes into this branch.  I did them through github, so I can't easily build.  Can you pull and rebuild, and let me know if there are errors?

kammitama5:
I got this error 
```
src/CodeWorld/Driver.hs:2124:36: error:
    • Variable not in scope: debugState :: DebugState
    • Perhaps you meant one of these:
        data constructor ‘DebugState’ (line 2011),
        data constructor ‘DebugStart’ (line 2018)
```

cdsmith:
Oops.  the `wrapper` in 2123 should be replaced by `wrapper@(debugState, _)`

kammitama5:
On the ```codeworld-api``` using ```cabal_install``` it builds with no errors. However, using ```./build.sh``` it aborts because of errors on 
```
2316 , 2568, 2586, 2587, 2598. 2606
(argument mismatches)
```

cdsmith:
Yes, this is expected.  All of this development happened on the GHCJS implementation.  The stand-alone GHC implementation has fallen behind.  That's okay; we can catch it up next!  Let's just get the core version working first.

codeworld-base/src/Internal/CodeWorld.hs
```diff
@@ -22,7 +22,9 @@ module Internal.CodeWorld
     , animationOf
     , activityOf
     , groupActivityOf
+    , CW.unsafeGroupActivityOf
```
cdsmith: This shouldn't be exported.  There's no need for the unsafe version to be exported from `codeworld-base`, because the monad bind operator isn't exported.  I'm not sure what you were trying to do, but there should be a different way.

codeworld-base/src/Internal/CodeWorld.hs
```diff
@@ -22,7 +22,9 @@ module Internal.CodeWorld
...
     , simulationOf
+    , CW.debugSimulationOf
```
cdsmith: You want to write a wrapper around this, instead of exporting it directly.  The version of this from `codeworld-api` has the type `a -> (Double -> a -> a) -> (a -> CW.Picture) -> IO ()`.  But the export here should have the different type:

    ([Number] -> a,
     (a, Number) -> a,
     a -> Picture)
    -> Program

So you need a wrapper.  It should look almost identical to the `simulationOf` wrapper that already exists below.

Same comment applies to `debugInteractionOf`, once you get there.

cdsmith: You'll need to re-export this from `Prelude`, too.

PR #698, opened 2018-07-25
History slider implementation

kammitama5:


cdsmith:
This looks awesome!

My biggest concern is that the placement of the two sliders doesn't make it clear what's going on.  I think we can do something subtle with arranging the parts to make this clearer.  Specifically, what about this layout?

    [A] [-----B-----] [C]      [D] [-----E-----] [F]

    A = PauseButton or PlayButton, depending on state
    B = SpeedSlider
    C = FastFowardButton (doesn't exist yet; increases speed)
    D = UndoButton, if applicable
    E = HistorySlider
    F = RedoButton or StepButton, depending on state

This would make sense to me, and I think be nicely discoverable.

cdsmith:
It would also help if we add an "x" to the end of the `SpeedSlider` label, so it reads "2x" instead o just "2".

PR #733, opened 2018-09-22
Move toolbar out of jQuery-layout and allow line-wrapping of buttons.

### LINKED ISSUE #732: Run button is off-screen when zooming on a small screen
When zooming in on a small screen, the Run button is pushed off to the right.  We'd like Run and Stop to pin themselves to the right margin, even if it means losing some of the buttons in the middle.

Request by @marcianx

marcianx:
Fixes #732 by allowing the toolbar to wrap and also taking it out of the jQuery-layout so that it can vertically size itself appropriately without javascript (since this resize sometimes fails to do so with jQuery-layout).
This PR also includes a minor fix to allow vertical scroll when the content overflows in the results pane.

One subtle glitch is that if you zoom in a LOT, then attempting to resize the result canvas to be smaller (drag mouse on the separator) sometimes catches in the middle and doesn't resize further. See, e.g., this GIF:
resize
When zooming back out, it behaves fine. WDYT?

Further notes:

1. One can implement lighter-weight resizable sections using the [CSS resize property](https://developer.mozilla.org/en-US/docs/Web/CSS/resize) that might make it easier to control the layout; however, this feature not yet supported by Edge.
2. This project might want to consider incorporating [a CSS auto-prefixer](https://github.com/postcss/autoprefixer) to avoid needing to manually add/maintain vendor prefixes.

### DIFF
--- a/web/blocks.html
+++ b/web/blocks.html
@@ -38,70 +38,70 @@
     </script>
   </head>
   <body>
-    <div id="nav" class="dropbox ui-layout-west">
-      <div class="nav_items">
-        <a id="newFolderButton" class="cw-button red" onclick="newFolder()">
-          <i class="mdi mdi-18px mdi-folder"></i>&nbsp; New Folder
-        </a>
-        <a id="newButton" class="cw-button red" onclick="newProject()">
-          <i class="mdi mdi-18px mdi-plus"></i>&nbsp; New
-        </a>
-        <a id="saveButton" class="cw-button blue" style="display:none" onclick="saveProject()">
-          <i class="mdi mdi-18px mdi-cloud-upload"></i>&nbsp; Save
-        </a>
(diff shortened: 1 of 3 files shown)


marcianx:
Updated commit to use the right email address to satisfy the CLA bot.

marcianx:
Further notes

3. Another feature that's *very* easy to implement is support for resizing only the code's font and no other part of the UI. WDYT?

marcianx:
Nevermind my last comment, I just discovered Ctrl +/-. You'd mentioned this before, but I think, I might might have not being paying attention.

cdsmith:
Yeah, it looks like you might just have to register that MutationObserver on the `.ui-layout-container`, instead of `document.body`.  It's listening for child elements to see when the cloned drag bar is added, and stop the running program from grabbing pointer events.

cdsmith:
Will push to production shortly.

PR #736, opened 2018-09-24
fix(install): initialize color variables for non color capable terminals

pulsation:
Fixes the following error : 
```
base.sh: line 52: setgreen: unbound variable
```
when running install on a non color capable terminal, for example when building a docker image.


cdsmith:
Thanks

PR #738, opened 2018-10-01
Add chessai as an honorary contributor

cdsmith:
Although the contribution was technically to GHC, chessai is helping to make it much easier to build and deploy CodeWorld by avoiding the need to patch GHC.

@chessai

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -13,6 +13,7 @@
 Fernando Alegre <alphalambda@fernando.alemor.org>
 Peter Becich <peterbecich@gmail.com>
 Joachim Breitner <mail@joachim-breitner.de>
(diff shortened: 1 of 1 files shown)


chessai:
Thanks!

chessai:
Unfortunate that it's not getting merged yet. Just saw https://ghc.haskell.org/trac/ghc/ticket/15702

cdsmith:
Yeah, I guess that makes sense, though.  Looks like I should just check whether there was a module specified in `-main-is`, and if so, only change the default exports when that's the current module being compiled.  I'm not quite fluent enough on GHC to know how to do that, but it doesn't seem hard.  Maybe I'll take a look at Haskathon this weekend.

PR #749, opened 2018-10-07
Avoid setting CSS height on body since sweetAlert2 overrides the height.

### LINKED ISSUE #748: Some modal dialogs collapses most of the screen behind them
Dialogs created by `sweetAlert2` cause the editor window to collapse to a height of zero behind the dialog window.  The bottom toolbar is still shown, but the entire top pane collapses.

The easiest way to reproduce is to just run `sweetAlert2('testing')` from the JavaScript console.

This seems to have started with https://github.com/google/codeworld/commit/a95a10b9553e072932f78f78c2b91934f5d73907 by @marcianx which moved the split pane to a subpane of the root.

marcianx:
Fixes #748.

Just adds a new `<div>` wrapping the layout content and moves `<body>`'s CSS into that.

### DIFF
--- a/web/blocks.html
+++ b/web/blocks.html
@@ -38,70 +38,72 @@
     </script>
   </head>
   <body>
-    <div class="ui-layout-container">
-      <div id="nav" class="dropbox ui-layout-west">
-        <div class="nav_items">
-          <a id="newFolderButton" class="cw-button red" onclick="newFolder()">
-            <i class="mdi mdi-18px mdi-folder"></i>&nbsp; New Folder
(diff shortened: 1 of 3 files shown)


cdsmith:
Thanks!  I tried to poke into this last week, but couldn't figure it out.

cdsmith:
Huh.  According to https://sweetalert2.github.io/#configuration, there's an option to disable setting 'height: auto' on the document body.  That said, though, there must be some reason it happens?  So I still like the solution with a nested div better.

Issue #10, opened 2014-06-25
Doc on hover


cdsmith:
When the user hovers the mouse over a standard library function, we should pop up a tool tip with basic type and documentation for that symbol.

nixorn:
Hi from BountySource. I see this issue listed here https://www.bountysource.com/teams/codeworld/bounties, can I try to solve this?

nixorn:
Is this issue still actual? Or any issues on bountysource?

cdsmith:
Hi.  Yes, this is still an actual issue, and you are welcome to work on it.  I'm not aware of anyone else doing any work on this.  Let me know if you have any trouble getting started.

You probably want this feature to use the same information as autocomplete, which is loaded here: https://github.com/google/codeworld/blob/master/web/js/codeworld_shared.js#L148 and the file being parsed is https://code.world/codeworld-base.txt .  Right now, the doc comments are ignored when reading that file, and only names and type signatures are pulled out.  I'd suggest starting with just that; a good first step would be to just display names and type signatures.  To get the docs as well, you'd want to extend the parsing at https://github.com/google/codeworld/blob/master/web/js/codeworld_shared.js#L247 to read and build up a string that's applied to the next symbol.

nixorn:
Thanks for explanation, I started.

nixorn:
Looks CodeMirror tracks only editor cursor, not mouse coordinates, and have no event like "token.mouseover()" from jquery. I have not figure out with CodeMirror completely yet, and in search of best way.

Now I see 2 options:
1. use CodeMirror cursor events to determine when some text become highlighted variable, and create hidden tooltip element for this token like [here](https://www.w3schools.com/howto/tryit.asp?filename=tryhow_css_tooltip). Css solution without extending CodeMirror defaults. 
2. write CodeMirror custom addon, which will handle hover events. I'm not sure about this, need to deeper understanding of CodeMirror internals for that. 

I did not start parsing yet, left it for dessert.

cdsmith:
Oh, it looks like that addon may already exist in https://www.npmjs.com/package/codemirror-extension  I don't know much about npm, so I'm not sure how to best incorporate that into the project.

cdsmith:
It looks like this is working.  Great!  I'll file a few minor bugs, but I think we can close this issue.  Thanks so much for the contribution.

PR #795, opened 2019-01-18
Resolve doc on hover

nixorn:
Resolve doc on hover #10

### DIFF
--- a/web/js/codeworld_shared.js
+++ b/web/js/codeworld_shared.js
@@ -70,6 +70,72 @@ var Alert = (() => {
   return mine;
 })();
 
+var hintBlacklist = [
+    // Symbols that only exist to implement RebindableSyntax or map to
+    // built-in Haskell types.
+    "Bool",
+    "IO",
+    "fail",
+    "fromCWText",
+    "fromDouble",
+    "fromInt",
+    "fromInteger",
+    "fromRational",
+    "fromString",
+    "ifThenElse",
+    "toCWText",
+    "toDouble",
+    "toInt",
+
+    // Deprecated exports.
+    "MousePress",
+    "MouseRelease",
+    "MouseMovement",
+    "MouseButton",
+    "LeftButton",
+    "RightButton",
+    "MiddleButton",
+    "Maybe",
+    "Nothing",
+    "Just",
+    "withDefault",
+    "hasValue",
+    "definitely",
+    "path",
+    "thickPath",
+    "text",
+    "styledText",
+
+    // Experimental exports.
+    "White",
+    "Black",
+    "Gray",
+    "Grey",
+    "Red",
+    "Orange",
+    "Yellow",
+    "Green",
+    "Blue",
+    "Purple",
+    "Pink",
+    "Brown",
+    "Pi",
+];
+
+var codeWorldBuiltinsDocs = {};
+
+function onHover(cm, data, node){
+    var result = null;
+    if (data && data.token && data.token.string != "") {
+        var token_name = data.token.string;
+        if (token_name && hintBlacklist.indexOf(token_name) == -1 && codeWorldBuiltinsDocs[token_name]){
+            result = codeWorldBuiltinsDocs[token_name];
+        }
+    }
+    return result;
+}
+
+// Hints and hover tooltips
 function registerStandardHints(successFunc)
 {
     function createHint(line, wordStart, wordEnd, cname) {
@@ -98,6 +164,21 @@ function registerStandardHints(successFunc)
             render: renderer,
             source: line
         };
+    }
+
+    function createHover(line, wordStart, wordEnd, doc, hint){
+        if (!hint){
+            hint = createHint(line, wordStart, wordEnd);
+        }
+        var annotation = document.createElement('div');
+        hint.render(annotation);
+        var description = document.createElement('div');
+        description.innerHTML = doc;
+        description.className = " hover-doc";
+        var docDiv = document.createElement('div');
+        docDiv.appendChild(annotation);
+        docDiv.appendChild(description);
+        return docDiv
     }
 
     // Add hint highlighting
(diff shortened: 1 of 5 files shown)


cdsmith:
Do you have a running server with this code?  If not, I can spin up my own to test with.

web/js/codeworld_shared.js
```diff
@@ -70,6 +70,72 @@ var Alert = (() => {
...
+    "Pi",
+];
+
+var codeWorldBuiltinsDocs = {};
+
+function onHover(cm, data, node){
+    var result = null;
+    if (data && data.token && data.token.string != "") {
+        var token_name = data.token.string;
+        if (token_name && hintBlacklist.indexOf(token_name) == -1 && codeWorldBuiltinsDocs[token_name]){
```
cdsmith: [nit] If you're going to test data.token.string for truthiness, why not do it above where you currently compare with ""?

web/js/codeworld_shared.js
```diff
@@ -70,6 +70,72 @@ var Alert = (() => {
...
+    "Purple",
+    "Pink",
+    "Brown",
+    "Pi",
+];
+
+var codeWorldBuiltinsDocs = {};
+
+function onHover(cm, data, node){
+    var result = null;
```
cdsmith: [nit] I think it would be cleaner to just add two returns, and not use the local var.

web/js/codeworld_shared.js
```diff
@@ -168,62 +249,10 @@ function registerStandardHints(successFunc)
...
-        "Orange",
-        "Yellow",
-        "Green",
-        "Blue",
-        "Purple",
-        "Pink",
-        "Brown",
-        "Pi",
-    ];
+    codeWorldBuiltinsDocs['main'] = createHover('main :: Program', 0, 4, 'Your program.');
```
cdsmith: I don't think it's important to add docs for deprecated code styles.  We can leave out `main`

web/js/codeworld_shared.js
```diff
@@ -168,62 +249,10 @@ function registerStandardHints(successFunc)
     codeworldKeywords['main'] = 'deprecated';
     codeworldKeywords['program'] = 'builtin';
 
-    lines = lines.sort().filter(function(item, pos, array) {
-        return !pos || item != array[pos - 1];
```
cdsmith: It looks like this was deduping the items.  I'm not 100% sure why it was necessary.  Do you know, and have you handled it a different way?

web/js/codeworld_shared.js
```diff
@@ -260,42 +287,48 @@ function registerStandardHints(successFunc)
...
 
-        var wordStart = 0;
-        if (line.startsWith("type ") || line.startsWith("data ")) {
-            wordStart += 5;
-
-            // Hide kind annotations.
-            var kindIndex = line.indexOf(" ::");
-            if (kindIndex != -1) {
-                line = line.substr(0, kindIndex);
+        if (line.startsWith("-- ")) {
```
cdsmith: To be technically correct, I think what you want is:

    if (line.startsWith("-- |")) {
      doc = ...
    } else if (doc != "" && line.startsWith("-- ")) {
      doc += ...
    }

web/js/codeworld_shared.js
```diff
@@ -260,42 +287,48 @@ function registerStandardHints(successFunc)
...
+            doc += line.replace(/(\-\- \| |\-\-   )/g, "") + "<br>";
```
cdsmith: I don't believe you want the `<br>` there.  I believe you just want `\n`.  Line wrapping in haddock comments is not significant except in preformatted blocks.

cdsmith:
A quick spot check reveals this in `codeworld-base.txt`:

    -- | <i>Warning: This is an experimental API.</i>

    -- | <i>Warning: This is an experimental API.</i>

    -- | <i>Warning: This is an experimental API.</i>

    -- | <i>Warning: This is an experimental API.</i>
    type Point = (Number, Number)
    translatedPoint :: (Point, Number, Number) -> Point
    rotatedPoint :: (Point, Number) -> Point

This seems to be a Haddock bug: those warnings are on pattern synonyms that occur at the end of the file.  But it's unfortunate, because one of them ends up getting attached to `Point` by accident.  I don't know that we can fix this right away, but if you can think of a workaround, that would be great.

cdsmith:
By the way, thank you for this awesome pull request!

nixorn:
Yes, I have deployed codeworld locally, so I have server with this code.
Will fix most of review points in a hour, except maybe deduplication of codeworld-base.txt content.

web/js/codeworld_shared.js
nixorn: True. Updating.

web/js/codeworld_shared.js
nixorn: Ok, done.

web/js/codeworld_shared.js
nixorn: Ok.

web/js/codeworld_shared.js
nixorn: Btw, what is "[nit]"?  Deleted `!= ""`

web/js/codeworld_shared.js
nixorn: No, I missed that. Will do this in separate commit.

web/js/codeworld_shared.js
nixorn: Work with `\n`, replacing.

cdsmith:
Great!  Yeah, I don't know that there's an answer for the pattern synonym issue.  What it comes down to is:

1. The version of haddock here skips pattern synonyms entirely.  (https://github.com/haskell/haddock/issues/946)
2. At some point, I added a workaround by appending them to the end of the file. (https://github.com/google/codeworld/blob/master/build.sh#L37)

But of course that workaround doesn't capture the haddock.  Even worse, the haddock is still emitted, but without the declaration.

Don't worry about this, actually.  In about a week, I plan to remove the WARNING pragmas on those pattern synonyms, so it's not worth fixing right now.  This coming summer, I intend to upgrade to GHC 8.6, and hopefully will pick up a new haddock at the same time with the bug fix, and also remove the manual workaround.  That will be the correct fix.

nixorn:
Should I care about deduplication of `codeworld-base.txt` in js then?

web/js/codeworld_shared.js
cdsmith: Oh, [nit] means I'm admittedly nitpicking.  Basically, I mean if you feel the original is better, it's not something I'd push back on.

cdsmith:
I don't know if dedup is necessary.  I think I'll merge this as is, play around with it, and fix any issues as they come up.

Issue #33, opened 2014-07-03
Auto-complete should use symbols defined in the current program


cdsmith:
Right now, auto-complete works against a fixed list of symbols from the standard library.  In addition to that, it should look in the current file for things that look like definitions, and add them to the auto-complete list.

Maaarcocr:
I would be interested in working on this feature for the Summer of Haskell. I would probably work on different others feature such as issue number 34: Auto-complete should be type-sensitive and issue number 129: Rainbow brackets. Would it be possible to further discuss about this possibility?

cdsmith:
@Maaarcocr Sure, tackling a cluster editor usability features would be a great proposal!  Feel free to get in touch with me at cdsmith@gmail.com to discuss any questions you have.

nixorn:
If it is actual, I can try.

cdsmith:
It looks like a rough version of this was added in https://github.com/google/codeworld/commit/c63fd5ccb6d6c1017e643568216a005312b5fd42 by a Summer of Haskell student two years ago.  I missed it because it was one commit in a long list.

Remaining issues:

1. It only works for symbols with top-level type signatures.  This is probably okay, since implementing type inference in the browser is less than ideal.
2. The newly added symbols are appended to the bottom of the autocomplete list, which makes this feature less discoverable than it could be.
3. User-defined symbols are not included in the newly implemented doc-on-hover feature.  It would be great to add them there as well.

Points 2 and 3 together indicate that there's room for keeping some kind of canonical dictionary of known terms that is updated as editing continues, and is used as a common source for both hover and autocomplete.

cdsmith:
@nixorn I noticed that you indicated you're working on this for [BountySource](https://www.bountysource.com/issues/2940977-auto-complete-should-use-symbols-defined-in-the-current-program).  As I mentioned above, @three implemented this in the summer of 2017.  I made a small improvement to sorting (which may have been broken by the doc-on-hover CL anyway, but I don't think so).  The core feature here already exists.  There are definitely improvements, but this gets messy in terms of the bounty.  I'd have to award the bounty to @three if he wants to claim it; otherwise, if the two of you agree to something else, I'm happy to go along.

cdsmith:
[...]

Issue #798, opened 2019-01-20
Better parsing for extracting autocomplete/hint options from source code


cdsmith:
There's some existing code for finding type declarations in the current module, and augmenting the autocomplete list with them.  @nixorn is extending that to his new doc-on-hover feature, as well.  However, the extraction code isn't all that great right now.  Specifically:

1. It only looks at one line at a time.  This generates incorrect results when the type declaration is wrapped across lines.
2. It mis-parses some syntax.  For example `pic1,pic2 :: Picture` will add a single item called `pic1,pic2`, when it should add separate items for both of the two pictures.
3. It would be nice to extend the parsing to recognize definition lines (like `f x = sqrt x + 1`).  Type inference isn't needed here, but I think it would be great if that gave `f` an incomplete type in autocomplete and docs, like `f :: ? -> ?`.  Of course, an explicit type signature should replace the incomplete definition with a complete one.
4. When extraction does succeed, it still leaves unnecessary formatting in place.  So if the original source code says:

       short                    :: Number
       reallyReallyVeryVeryLong :: Number

   the extracted doc for `short` will have `short                    :: Number`, even though those extra spaces aren't sensible in the new context.

One must probably resist the temptation to completely reimplement the Haskell parser in JavaScript, but I think enough special cases could go a long way.

cdsmith:
Just to record this, I did think about the possibility of using the GHC API on the server to extract type information.  This would support any syntax corner case with no effort, and would even give fully inferred types, which is great.  However, there are two reasons this wouldn't work so well:

1. It's a lot of work.  Step 1, for instance, would be getting `codeworld-base` to build with plain GHC instead of GHCJS in the first place.
2. It's potentially expensive for the server.  That means rate limiting, which means docs and autocomplete can get very out of date.
3. Worst of all, it only works when the whole module is syntactically correct.  A single syntax error on the other side of the module would break autocomplete everywhere.  I wouldn't consider this acceptable.

As a result, this is an idea that would, at best, be an occasional side benefit on top of the client-side implementation that we'd need anyway.  Then the only huge benefit is the inferred types, since the rest of this should be handled on the client anyway.  And inferred types are nice, but I'm actually kind of okay with students seeing question-marks in their autocomplete, and telling them they should just write a type declaration to see the whole type.

So I think this should just be rejected, and an implementation on the client should be the focus.

cdsmith:
More examples to fix:

    {- ==================
    Some block comment
    ================== -}

[...]

PR #799, opened 2019-01-23
Autocomplete using symbols from current programm + Auto-complete should include documentation

### LINKED ISSUE #135: Auto-complete should include documentation
The autocomplete window include type annotations, but it should include documentation as well.  Possibly the one-line summary for all entries, and possibly a more complete doc window for the currently selected auto-complete entry.

nixorn:
Closes #135 #33


nixorn:
To be honest I confused with css updates from last commits. I'm not sure if this request will broke some of your work, I'm sorry if it will.

web/js/codeworld_shared.js
```diff
@@ -122,13 +122,111 @@ var hintBlacklist = [
...
+    if (keywordData.doc) {
+        var docElem = document.createElement('div');
+        docElem.innerHTML = keywordData.doc;
+        docElem.className = "hover-doc";
+        elem.appendChild(docElem)
+    };
+    return elem;
+};
+
+function applyCodemirrorChange(line, change) {
```
cdsmith: This is worrisome.  Does this mean that CodeMirror will silently ignore any kind of edit except for insertions and deletions?  Skimming the documentation, it looks like at least "+move" is an option, and others besides.  I'm pretty convinced this is the wrong way to go, as it's likely to be a minefield that will break with every editor feature you haven't quite tested yet.

What's the purpose behind ignoring `setValue` and `complete`?

web/js/codeworld_shared.js
```diff
@@ -122,13 +122,111 @@ var hintBlacklist = [
...
+// codeWorldDocs = {
+//   codeWorldLogo: {
+//     annotation: "Picture",
+//     doc: "The CodeWorld logo."
+//   }
+// }
+
+var codeWorldDocs = {};
+
+function renderHover(keyword) {
```
cdsmith: This looks like it only handles lines of the form `foo :: Type` or `foo`.  What about `(*#^) :: Type`, or `data Foo`, or `type Foo = Bar`?

web/js/codeworld_shared.js
```diff
@@ -122,13 +122,111 @@ var hintBlacklist = [
...
+        console.log("Event \"" + change.origin + "\" not handled.");
+        return;
+    }
+}
+
+function getWordAndAnnotation(line) {
+    var annotation = null,
+        word = null;
+    // there is top-level assignment
+    if (/^\S+\s*=/.test(line)) {
```
cdsmith: Will this match lines like `f(x, y) =`?  (Note the space after the comma.)  I don't think so, since the `=` doesn't immediately follow the first whitespace range.

web/js/codeworld_shared.js
```diff
@@ -122,13 +122,111 @@ var hintBlacklist = [
...
+
+    before_data = getWordAndAnnotation(before)
+    after_data = getWordAndAnnotation(after)
+
+    // user updated name/annotaion of some variable
+    // updating hovers/hints information
+    if (!(before_data.word === after_data.word
+          && before_data.annotation === after_data.annotation
+          // We don't want to override builtins.
+          // Marker of builtin variable - it have docs.
```
cdsmith: I'm not sure about this invariant that builtin symbols have docs.  It may be true today (though not always), but it could change in the future if we start to parse docs in the current module.  What about adding a boolean field to the object instead?

web/js/codeworld_shared.js
```diff
@@ -122,13 +122,111 @@ var hintBlacklist = [
...
+        word = splitted[0].trim()
+        annotation = splitted[1].trim()
+    };
+    return {
+        word: word,
+        annotation: annotation
+    }
+}
+
+function updateDocs(cm, change) {
```
cdsmith: I am skeptical of this whole approach.  I'd be more comfortable with something like setting a timer when the document changes, and re-parsing the whole document when the timer expires.  My fear is that this is likely to get out of sync, leaving us in a situation where the docs or autocomplete are wrong, and the only way to fix it is to reload the document.

cdsmith: I suspect this is also the reason you've chosen to ignore lots of edit types above, which is more evidence it's probably not the right way to go.

web/js/codeworld_shared.js
```diff
@@ -225,31 +250,56 @@ function registerStandardHints(successFunc)
...
+
+            CodeMirror.on(data, 'close', deleteOldHintDocs);
+            CodeMirror.on(data, 'pick', deleteOldHintDocs);
+
+            // Tracking of hint selection
+            CodeMirror.on(
+                data, 'select',
+                function (selection, elem) {
+                    var codeWordInfo = codeWorldDocs[selection],
+                        hihtsWidget = elem.parentElement,
```
cdsmith: Misspelling.  It looks like this variable isn't used anyway.

install.sh
```diff
@@ -252,7 +252,7 @@ run $BUILD  rm -rf downloads
 run $BUILD            git clone https://github.com/codemirror/CodeMirror.git
 run $BUILD/CodeMirror git checkout tags/5.25.2
 run $BUILD/CodeMirror npm install
-run $BUILD/CodeMirror npm install -s uglify-js codemirror-extension
+run $BUILD/CodeMirror npm install -s uglify-js codemirror-extension@0.2.0
```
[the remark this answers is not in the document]
nixorn: Just to freeze working version. This package looks like dead, and chance of new version is low but who knows...

web/js/codeworld_shared.js
nixorn: Ok, will add.

web/js/codeworld_shared.js
nixorn: Thanks for correction, will extend parsing rules.

web/js/codeworld_shared.js
nixorn: Will extend rules.

web/js/codeworld_shared.js
nixorn: In this case this mean that CodeMirror will ignore all except insertions and deletions in [beforeChange handler](https://github.com/google/codeworld/pull/799/files#diff-0b4870211bd48634500a9dccbf7798dfR173). 
But anyway, `applyCodemirrorChange` exactly not finished yet. If it will be used in other places it could break something. `complete` ignored because have no use (I could be wrong) in updating doc dictionary. If some term could be completed - it already in documentation.

About `setValue`. `setValue` happens on loading project, so yes, it is error. All keywords in fresh loaded project will not be parsed.

Also, `applyCodeMirrorChange` works only per line now, I guess there is actions which could touch  more than one.

 I will think about all this.

web/js/codeworld_shared.js
nixorn: Got it. Reasonable.
I'll try to rewrite from event-oriented approach to single parsing function + timer.

nixorn: Done. Take a look please.

web/js/codeworld_shared.js
nixorn: Thx, deleted.

cdsmith:
Thanks for the update.  I'm pretty busy today at work, but I'll try to take a look tonight when I have time.

web/js/codeworld_shared.js
```diff
@@ -122,95 +122,142 @@ var hintBlacklist = [
...
+//   codeWorldLogo: {
+//     annotation: "Picture",
+//     doc: "The CodeWorld logo.",
+//     builtin: true // codeWorld definitions, not from current project
+//   }
+// }
+
+var codeWorldDocs = {};
+
+function codeToDoc() {
```
cdsmith: This currently adds any new symbols defined in the source code, but it doesn't remove any old ones.  That's a problem.

Thinking about this, actually, I think the right thing to do is something like this:

    var codeWorldBuiltinDocs = {};
    var codeWorldDocs = {};

    function codeToDoc() {
      codeWorldDocs = Object.assign({}, codeWorldBuiltinDocs);
      ... // add symbols from the current module
    }

That way you start with a fresh copy of the builtins each time, and don't risk leaving around whatever weird stuff you might have picked up from the previous parse of the loaded code.

This way, you don't really need the `builtin` property.

web/js/codeworld.js
```diff
@@ -171,6 +171,9 @@ function initCodeworld() {
 
     window.codeworldEditor.on('changes', window.updateUI);
 
+    setTimeout(codeToDoc, 5000);
+    setInterval(codeToDoc, 30000);
```
cdsmith: I wonder if you should do something like:

    window.reparseTimeoutId = null;
    window.codeworldEditor.on("change", () => {
      if (window.reparseTimeoutId) clearTimeout(window.reparseTimeoutId);
      window.reparseTimeoutId = setTimeout(codeToDoc, 3000);
    });

That way, the symbol list would be relatively up-to-date (max 3 seconds old), but you wouldn't waste time constantly reparsing when the document isn't modified, and you also wouldn't risk typing being choppy because of constant reparses while the user is typing.  I think that's the right trade-off here.

web/js/codeworld.js
```diff
@@ -171,6 +171,9 @@ function initCodeworld() {
 
     window.codeworldEditor.on('changes', window.updateUI);
 
+    setTimeout(codeToDoc, 5000);
```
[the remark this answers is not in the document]
nixorn: It its forgotten wrong way to wait until code loaded to editor. Going to move this call to [setCode](https://github.com/google/codeworld/blob/master/web/js/codeworld.js#L629) function.

nixorn:
One more update :smile:

web/js/codeworld_shared.js
```diff
@@ -122,95 +122,161 @@ var hintBlacklist = [
...
+            parseResults[word] = {
+                symbolStart: getWordStart(word, line),
+                symbolEnd: getWordEnd(word, line)
+            };
+        }
+        // data Foo
+        else if (/^data\s.+/.test(line)) {
+            word = line.split(" ")[1];
+            parseResults[word] = {
+                declaration: line.slice(0, getWordStart(word, line)),
```
cdsmith: did you mean `getWordEnd`?

web/js/codeworld_shared.js
```diff
@@ -122,95 +122,161 @@ var hintBlacklist = [
...
-        };
-    }
-
-    function createHover(line, wordStart, wordEnd, doc, hint){
-        if (!hint){
-            hint = createHint(line, wordStart, wordEnd);
+        // type Foo = Bar
+        else if (/^type\s.+/.test(line)) {
+            var splitted = line.split("=");
+            word = splitted[0].trim().split(" ").slice(-1)[0];
```
cdsmith: Same as above: use capturing groups to extract the name.  This won't work for `type Foo=Bar` for example.  I won't keep repeating this; hopefully you get the idea.

web/js/codeworld_shared.js
```diff
@@ -122,95 +122,142 @@ var hintBlacklist = [
...
+//   }
+// }
+
+var codeWorldDocs = {};
+
+function codeToDoc() {
+    var lines = window.codeworldEditor.getValue().split('\n'),
+        word = null,
+        annotation = null;
+    lines.forEach(function(line) {
```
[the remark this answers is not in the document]
cdsmith: Okay to skip this, but let's file a new bug to track the cleanup.

web/js/codeworld_shared.js
```diff
@@ -122,95 +122,161 @@ var hintBlacklist = [
...
 
-        function renderer(elem, data, cur) {
-            if (wordStart > 0) {
-                elem.appendChild(document.createTextNode(line.slice(0, wordStart)));
+    lines.forEach(function(line) {
+        // f(x, y) =
+        if (/^\w+\(.*/.test(line)) {
+            word = line.split("(")[0];
+            parseResults[word] = {
+                symbolStart: getWordStart(word, line),
```
cdsmith: What does it mean to populate `symbolStart` and `symbolEnd` without `declaration`?  I thought those were offsets into `declaration`.

web/js/codeworld_shared.js
```diff
@@ -122,95 +122,161 @@ var hintBlacklist = [
...
 
-        if (doc !== "") {
-            var description = document.createElement('div');
-            description.innerHTML = doc;
-            description.className = " hover-doc";
-            docDiv.appendChild(description);
-        }
+    if (!codeWorldSymbols[keyword] // no entry
+        || // no docs or declaration - nothing to show on hover
+        !(codeWorldSymbols[keyword].doc
```
cdsmith: Why not assume that `declaration` will always be present, and when you're parsing, fill it in with just the symbol itself if you can't do any better?  That simplifies this logic a lot.  In fact, there's already code later that assumes `declaration` is populated.

web/js/codeworld_shared.js
```diff
@@ -225,31 +291,58 @@ function registerStandardHints(successFunc)
         }
 
         var found = [];
-
+        var hints = Object.keys(codeWorldSymbols)
```
cdsmith: Did you mean `codeWorldBuiltinSymbols`?

It's an interesting idea to highlight symbols defined in the current module, as well.  But I'm not sure this accomplishes that, because you'd have to re-register the `hint` helper every time the code changes, and you don't do that, right?

web/js/codeworld_shared.js
```diff
@@ -225,31 +291,58 @@ function registerStandardHints(successFunc)
...
+            function deleteOldHintDocs(){
+                $(".hint-description").remove()
+            }
+
+            CodeMirror.on(data, 'close', deleteOldHintDocs);
+            CodeMirror.on(data, 'pick', deleteOldHintDocs);
+
+            // Tracking of hint selection
+            CodeMirror.on(
+                data, 'select',
```
cdsmith: I can't find any clear documentation on this form of `on`.  Everything I'm looking at uses the first parameter to say which CodeMirror instance you're adding the listener to.  What does it mean to pass `data` there?

nixorn: `'select'`  is event from [hint/show-hint.js addon](https://codemirror.net/doc/manual.html#addon_show-hint).
"The following events will be fired on the completions object during completion: " 
Data here is "completions object", it is not obvious from documentation. I understood how it works after reading https://github.com/codemirror/CodeMirror/issues/3092 .

web/js/codeworld_shared.js
nixorn: It works. Don't need to re-register helper, logic don't changes. Every time user call `hint` helper make new completions source (`data`) from `codeWorldSymbols`.
Probably, code from `var hints = Object.keys(codeWorldSymbols)` with getting defined symbols and sorting of found completions should run on edit, not on hint.
image

web/js/codeworld_shared.js
nixorn: You are right, no meaning. Will delete.

web/js/codeworld_shared.js
nixorn: Yes, thank you!

web/js/codeworld_shared.js
nixorn: Done.

web/js/codeworld_shared.js
```diff
@@ -122,95 +122,161 @@ var hintBlacklist = [
...
+        else if (/^\S+\s*=/.test(line)) {
+            word = line.split("=")[0].trim();
+            parseResults[word] = {
+                symbolStart: getWordStart(word, line),
+                symbolEnd: getWordEnd(word, line)
+            };
+        }
+        // data Foo
+        else if (/^data\s.+/.test(line)) {
+            word = line.split(" ")[1];
```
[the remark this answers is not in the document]
nixorn: Done.

web/js/codeworld.js
```diff
@@ -170,6 +170,11 @@ function initCodeworld() {
     };
 
     window.codeworldEditor.on('changes', window.updateUI);
+    window.reparseTimeoutId = null;
+    window.codeworldEditor.on("change", () => {
```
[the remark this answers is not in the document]
nixorn: Ok.

web/css/codeworld.css
```diff
@@ -416,3 +416,18 @@ body {
 .req-details {
   color: #555;
 }
+.hint-word {
```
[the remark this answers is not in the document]
nixorn: ok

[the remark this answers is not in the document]
nixorn: Done

web/js/codeworld.js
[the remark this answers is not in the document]
nixorn: Done

web/css/codeworld-cm.css
```diff
@@ -85,26 +85,27 @@ li.CodeMirror-hint-active {
     color: #555;
 }
 
-.CodeMirror-hover-tooltip {
+.CodeMirror-hover-tooltip, .hint-description {
     background-color: #ffe;
-
     border-radius: 0.5em;
-    border: 1px solid #888;
+    border: 0.05em solid #888;
```
cdsmith: I still like `1px` for borders, as they look cleaner that way.  It's just significant distances that are better measured in `em`.

web/js/codeworld.js
```diff
@@ -169,7 +169,12 @@ function initCodeworld() {
         }
     };
 
-    window.codeworldEditor.on('changes', window.updateUI);
+    window.reparseTimeoutId = null;
+    window.codeworldEditor.on("changes", () => {
+        if (window.reparseTimeoutId) clearTimeout(window.reparseTimeoutId);
+        window.reparseTimeoutId = setTimeout(parseSymbolsFromCurrentCode, 3000);
+        window.updateUI;
```
cdsmith: This needs parentheses to call the function.

web/js/codeworld_shared.js
```diff
@@ -132,95 +132,157 @@ var hintBlacklist = [
...
-            if (wordStart > 0) {
-                elem.appendChild(document.createTextNode(line.slice(0, wordStart)));
+    lines.forEach(function(line) {
+        // f(x, y) =
+        if (/^\w+\(.*/.test(line)) {
+            word = line.split("(")[0].trim();
+            parseResults[word] = {};
+        }
+        // foo =
+        else if (/^\S+\s*=/.test(line)) {
```
cdsmith: There are definitely weird cases that will accidentally match here.  For instance:

    {- ==================
    Some block comment
    ================== -}

The first line will match and add a hint for "{-"

Also:

    a*&*b = a + 2 * b

This is actually a definition of the `*&*` operator, but will be misparsed as a definition for the symbol `a*&*b`.

I think this is okay for the initial merge, but should definitely file a bug (probably the same one as in the earlier comment) to replace this with an honest-to-goodness parser.  At the same time, we should make it able to recognize definitions that are split across line breaks, like

    f(x, y)
      = x + y

So let's leave this undone for now, but make sure to track the future work.

web/js/codeworld_shared.js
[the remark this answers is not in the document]
cdsmith: It turns out that #798 is a good place for this live.  I'll use that bug to track this work.

web/js/codeworld_shared.js
cdsmith: Will add this to #798

web/js/codeworld_shared.js
```diff
@@ -132,95 +132,157 @@ var hintBlacklist = [
...
+        if (hintBlacklist.indexOf(token_name) == -1){
+            return renderHover(token_name);
+        }
     }
+}
 
+// Hints and hover tooltips
+function registerStandardHints(successFunc)
+{
+    // Add hint highlighting
```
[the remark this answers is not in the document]
nixorn: Right, deleting.

web/css/codeworld-cm.css
nixorn: ok

web/js/codeworld.js
nixorn: done

PR #802, opened 2019-01-27
Improve the physics accuracy of the "bounce" example

josephcsible:


cdsmith:
Do you think this is an improvement?  Actually, if I were to make a change in the examples, it would be to simplify them, not make them more complex!  I fear that this moves this example from the realm of what's explainable to middle school students, into stuff that requires too much math for them.  It would be a shame if they concluded that physics is just complex pseudo-magic that is too hard to understand.

(Granted, some of the existing examples are worse.  But right now, this is one of the better ones.)

What about this alternative instead:

    fine :: ((a, Event) -> a, Number) -> ((a, Event) -> a)
    fine(f, max_dt) = ff
      where ff(x, TimePassing(dt))
              | dt > max_dt = ff(f(x, max_dt), TimePassing(dt - max_dt))
              | otherwise   = f(x, dt)
            ff(x, other) = f(x, other)

    program = activityOf(initial, fine(change, 0.01), picture)

I've used this `fine` combinator in a number of examples in the past where I want to minimize simulation error from large time steps.

josephcsible:
Good point about complexity. I completely threw out my old bounce changes and replaced them with a different approach. Instead of forcing the ball to always be completely inside the arena at the expense of potential+kinetic energy being inaccurate, it now keeps potential+kinetic energy completely correct at the expense of the ball sometimes slightly clipping over the edge of the arena. I consider this a good switch because the energy error accumulates over time, whereas the position error is completely nullified in the next time step.

I also added the `fine` function you mentioned to minimize the effect of the remaining error. I made a few minor changes to it, one to make it type-check and another for conciseness.

As for my first change (tweaking how gravity affects position), I feel that it actually makes it somewhat easier to understand, since we're now using one of the basic equations of motion exactly as they'd appear in a textbook, rather than trying to do a linear approximation of one.

cdsmith:
Thanks, this looks great.

PR #805, opened 2019-01-29
Use an exact value for translation instead of a badly rounded one

josephcsible:
Before:
before

After:
after


cdsmith:
Thanks again.

PR #807, opened 2019-01-31
Show CodeMirror compress error

josephcsible:
If this fails, it writes the error to stdout instead of stderr.
Detect this and print the output if this happens.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -78,6 +78,7 @@ function build_codemirror {
   exitcode=$?
   if [ $exitcode -ne 0 ]; then
     cat codemirror-compressed.js
+    rm codemirror-compressed.js
   fi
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -75,6 +75,11 @@ function build_codemirror {
                text-hover \
     --local node_modules/uglify-js/bin/uglifyjs \
     > codemirror-compressed.js
+  exitcode=$?
+  if [ $exitcode -ne 0 ]; then
+    cat codemirror-compressed.js
```
cdsmith: Nice catch.  Can you also delete `codemirror-compressed.js` in this case?  No point in leaving it around if it's not valid JavaScript anyway.

PR #809, opened 2019-02-02
Make various improvements to the Asteroids step function

josephcsible:
* Fix Asteroids last score sometimes being higher than max score
* Simplify energy calculation
* Be nicer about thrusting when out of energy


cdsmith:
@josephcsible Reverted the energy regeneration change, but left the others.  With the regeneration change, I feel it became less clear what the energy bar meant, since you'd still get thrust (albeit only half as much) after it was gone.

josephcsible:
@cdsmith Okay, here's my new solution for the energy problem: once the player runs out of energy, thrust goes "on cooldown", which means that even once there is energy again, thrust will remain off until either the energy bar fills all the way back up, or until the player releases and re-presses the Up key. This should fix the original problem without causing the confusion my original fix did.

Issue #812, opened 2019-02-05
Variable-sized brackets and bold text break character alignment in Firefox


josephcsible:
**Shared Program**: https://code.world/#PL1P3DfDlbKvMjslV59Pugg

**Description**

Consider lines 10 and 11 of the above program:
```haskell
change(w, KeyRelease("Left")) = w { leftReleases = leftReleases(w) + 1 }
change(w, _)                  = w
```
In Chrome, they look fine:
![image](https://user-images.githubusercontent.com/16061266/52250169-874a3800-28c4-11e9-9a16-156819ecb69d.png)

But in Firefox, the alignment is wrong:
![image](https://user-images.githubusercontent.com/16061266/52250197-9df08f00-28c4-11e9-8b04-dfb9e0082ebd.png)

Note the `= w` on line 11 is too far to the left. The same number of characters are there when you count, and it copies and pastes right, so it's just a display bug.

cdsmith:
Yes, this isn't surprising.  It's an artifact of the way brace highlighting is done.  Outer braces are rendered larger with a `letter-spacing` CSS property to try to compensate, but the exact right value for `letter-spacing` was chosen empirically.  See https://github.com/google/codeworld/blob/master/web/css/codeworld-cm.css#L38

This is going to be tough to fix, I'm afraid.  If we're lucky, maybe there's a better CSS unit that behaves consistently, but perhaps there's a way to do it by measuring the right letter-spacing value at runtime and modifying the CSS rules at runtime based on that measurement.

josephcsible:
I investigated this some more. It turns out that the font-size and letter-spacing have a little bit to do with it, but they're not responsible for most of it. (#820 fixes this bit, though.) The major culprit is that in Firefox, the Source Code Pro font takes up more horizontal space per character when it's bold. This doesn't happen with the standard "monospace" font, and it doesn't happen at all in Chrome. I'm not sure what's causing this, or what the best way to fix it is. I'll keep looking though.

josephcsible:
Okay, I see what's going on. The problem is that we're not loading the bold version of Source Code Pro, so the browsers are making one up themselves, and they're each doing it differently. Another PR coming up.

josephcsible:
Result of both changes in Chrome:
![image](https://user-images.githubusercontent.com/16061266/52315952-49621800-2987-11e9-8e0f-0f7cec0c1fe1.png)

And in Firefox:
[...]

PR #820, opened 2019-02-06
Use CSS transforms instead of playing with font-size for brackets

josephcsible:
This fixes part of #812, and also fixes vertical line spacing changing
due to parentheses.


cdsmith:
So, I'm curious, if there a way to make this actually change the line height, if I wanted it?  This is fine for now, but if I can get the brackets to resize based on the depth of nesting *inside* instead of distance from the outside, then I might want to ramp up the size a lot more for deeply nested parentheses, and then we'd probably want the line height to change.

I'm okay with this for now given the small change in sizes involved, but might have to revert it when I figure that out.

josephcsible:
The `line-height` property should do that. It changes the line spacing without changing the shape or size of any of the glyphs. (But I personally find it jarring when the line height changes due to characters added to it. Perhaps a better option would be to just bump up line heights throughout the code editor, just so it's consistent.)

PR #821, opened 2019-02-06
Load the actual Source Code Pro Bold font

### LINKED ISSUE #812: Variable-sized brackets and bold text break character alignment in Firefox

josephcsible:
Currently, we only load the normal Source Code Pro font, but we use the
bold version in some places anyway. This results in browsers making up
a bold variant, which is done inconsistently and often with bad metrics.
Load the actual bold variant instead. Fixes #812.


cdsmith:
Thanks for the fix.

PR #823, opened 2019-02-06
Resolve Better sync code editor with the running program

### LINKED ISSUE #133: Better sync code editor with the running program
One source of occasional student confusion is that they have edited their code, but are still running an old version of the program in the sidebar.  There are a few possibilities for fixing this:
- Remember the CodeMirror revision that's running, and grey out the run window, or otherwise mark it as obsolete, when the current revision is different from the running one.
- Automatically rebuild and restart the program on a timer of some sort when there's a change.
- Something more ambitious?  Like hot-plugging changes into a running program without losing the state?

The latter two options are worth being cautious about.  I've built a prototype that continually rebuilds and reruns user code in the past, and it's vulnerable to the weakness that partially typed student code can sometimes compile and go into an infinite loop, which can make browsers suddenly lock up or become intolerably slow.  (e.g., imagine a user is typing `[1..10]`, but the rebuild timer hits when they've written `[1..]`)

At least the first option should be easy enough, though.  Just need a good UI for it...

nixorn:
Closes #133

### DIFF
--- a/web/env.html
+++ b/web/env.html
@@ -82,6 +82,7 @@
         <form id="source" class="dropbox ui-layout-center"><textarea id="editor" placeholder="Type your code here..."></textarea></form>
 
         <div id="result" style="display:none" class="ui-layout-east">
+          <div id="obsolete-code-alert" class="obsolete-code-alert-fadeout"></div>
           <iframe id="runner" class="dropbox"></iframe>
           <pre id="message" class="dropbox"></pre>
         </div>
--- a/web/blocks.html
+++ b/web/blocks.html
@@ -75,6 +75,7 @@
         <div class="dropbox ui-layout-center" id="blocklyDiv" style="height: 84%; width:100%;"> </div>
 
         <div id="result" style="display:none" class="ui-layout-east">
+          <div id="obsolete-code-alert" class="obsolete-code-alert-fadeout"></div>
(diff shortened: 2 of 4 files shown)


nixorn:
I see iframe with `runner` id in two files: [blocks.html](https://github.com/google/codeworld/blob/master/web/blocks.html) and [env.html](https://github.com/google/codeworld/blob/master/web/env.html).

I added fading div into both. Looks env.html is used for haskell and codeWorld modes, and my changes works for these. But I can not test blocks mode, because did not figure out how block mode works.

web/env.html
```diff
@@ -82,6 +82,7 @@
         <form id="source" class="dropbox ui-layout-center"><textarea id="editor" placeholder="Type your code here..."></textarea></form>
 
         <div id="result" style="display:none" class="ui-layout-east">
+          <div id="obsolete-code-alert" class="obsolete-code-alert-fadeout"></div>
```
cdsmith: Should there be text on this?  Something like: "Your code has been modified.  Click Run to update the  running program."

cdsmith:
Thanks!  This looks good; I just suggested a change to add some text and explain what's going on.

cdsmith:
I'll just go ahead and merge now, and tweak the design later.

PR #836, opened 2019-02-12
Add position support for doc popup.

### LINKED ISSUE #87: Guide loses current position when it's closed

nixorn:
Closes #87

### DIFF
--- a/web/js/codeworld.js
+++ b/web/js/codeworld.js
@@ -675,6 +675,15 @@ function help() {
         allowEscapeKey: true,
         allowOutsideClick: true,
         showConfirmButton: false,
+    }, function(confirmed){
+        // Kind of onClose event for sweetalert. Function get single bool argument -
+        // true if confirmed, false if not.
+        if (confirmed){
+            console.log("Should never happen. Doc sweetalert have no confirm option.")
+        } else {
+            var docIframe = document.getElementById("doc");
+            docIframe.contentWindow.savePosition();
+        }
     });
 }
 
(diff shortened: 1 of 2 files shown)


web/js/codeworld.js
```diff
@@ -675,6 +675,15 @@ function help() {
         allowEscapeKey: true,
         allowOutsideClick: true,
         showConfirmButton: false,
+    }, function(confirmed){
+        // Kind of onClose event for sweetalert. Function get single bool argument -
+        // true if confirmed, false if not.
+        if (confirmed){
+            console.log("Should never happen. Doc sweetalert have no confirm option.")
```
cdsmith: I don't think we'd care if this did happen, such as in a future redesign of the UI.  So how about not bothering to check, and just storing the position anyway?

web/js/codeworld.js
```diff
@@ -675,6 +675,15 @@ function help() {
         allowEscapeKey: true,
         allowOutsideClick: true,
         showConfirmButton: false,
+    }, function(confirmed){
```
cdsmith: Please run the format.sh script that I added in the root codeworld directory.  It reformats to a consistent style everywhere, and updates some coding style things... for example, using arrow functions instead of lambdas, and let instead of var.

web/js/codeworld.js
nixorn: Done

web/js/codeworld.js
nixorn: Done

PR #843, opened 2019-02-20
Replace compile errors to deffered runtime and lint support for this

### LINKED ISSUE #137: Deferred errors for CodeWorld
It would be great if CodeWorld could defer some kinds of errors,  The deferred type error extension is one example.  It would actually be great to defer some kinds of syntax errors, as well.

We'd want to be careful with messages.  When a deferred error occurs at runtime, we would want to match it up with the compile-time warning, and just emphasize or promote the warnings into errors.

nixorn:
Closes #137


web/js/codeworld_shared.js
```diff
@@ -1203,4 +1203,19 @@ function printMessage(type, message) {
...
     outputDiv.scrollTop = outputDiv.scrollHeight;
+
+    if (isRuntime && type === "error") {
+        parseCompileErrors(message).forEach(
+            (error) => {
+                runtimeErrors[error.fullText] = error;
+            }
+        )
+        let cm = window.codeworldEditor;
+        cm.performLint();
```
cdsmith: Will this send another HTTP request to the server?  I don't think we want to do that.  Instead, I believe we want to cache the lint results for the current generation of code in a global variable.  The linter should check the generation against the latest one, and if it's the same, it will just work with the cached results.  Only if the source has been changed since then would we send a new request to the server.

web/js/codeworld.js
```diff
@@ -200,8 +202,15 @@ function initCodeworld() {
 
                         if (request.status == 400 || request.status ==
                             200) {
-                            callback(parseCompileErrors(request
-                                .responseText))
+                            let errors = parseCompileErrors(request.responseText);
+                            // Runtime errors is priority source compared to
```
cdsmith: Runtime errors only matter if the current source generation is the same as the running source generation.  The running source generation is already tracked, so this is just a matter of doing a comparison.

web/js/codeworld_shared.js
```diff
@@ -1203,4 +1203,19 @@ function printMessage(type, message) {
     }
 
     outputDiv.scrollTop = outputDiv.scrollHeight;
+
+    if (isRuntime && type === "error") {
```
cdsmith: What about updating the error in the output window?  That's actually more important than updating the linter, I think.

nixorn: I think output window is kind of terminal output. So updating of already printed line is weird for me. 

We can update existing line, or delete old one and print new at the end. What is better?

web/js/codeworld_shared.js
nixorn: It is good remark. Will fix.

web/js/codeworld.js
nixorn: Thanks for explanation.

web/js/codeworld_shared.js
cdsmith: I think we should update the existing line.  I realize that seems weird if you think of the output window as a terminal, but:
1. I don't think my students really do think of it that way.  In fact, I don't think they usually have any notion of what a terminal is.
2. As more and more errors are deferred, we end up in a weird position where most non-parse error messages are printed twice.  That's obviously not what you would want.

In this case, you have essentially two of the same message, differing only in whether it's a warning or an error.  It makes sense to just change the warning to an error, rather than appending it at the end.

web/js/codeworld_shared.js
```diff
@@ -1203,4 +1203,19 @@ function printMessage(type, message) {
...
+            }
+        )
+        let cm = window.codeworldEditor;
+        cm.performLint();
+    }
+}
+
+function matchWarningToDeferredError(msg){
+    return msg.replace("warning", "error")
+        + "\n(deferred type error)";
```
cdsmith: I didn't realize that GHC adds this "(deferred type error)" thing.  I think we should strip it off, since it won't be meaningful to students who will just come to expect the new behavior.

I do wonder, though, whether we ought to also promote these deferred errors to full errors in the UI.  I mean, maybe the program should still run, but we should display them as errors from the start?  In that case, we could look for "(deferred type error)", strip it off, but also change "warning" to "error" at the same time.

If you do it that way, then there's no need to go back later and fix up the errors when the program crashes.  You'd just filter out the runtime version of the deferred error, or maybe replace it with a brief message like "Giving up because of the error at program.hs:10:42".

nixorn: Ok, got it.

nixorn: There is some confusion. When it is warning it have no "(deferred type error)" postfix, only when it is already runtime error. So I have no quick way to understand if this warning is future error.

cdsmith: Okay, I see that now.  It looks like the only way to recognize these deferred warnings is actually before the error sanitizer runs.  Here's what I get from GHC itself:

    Test.hs:2:21: warning: [-Wdeferred-type-errors]
        • No instance for (Fractional Int) arising from the literal ‘4.5’
        • In the first argument of ‘take’, namely ‘4.5’
          In the second argument of ‘($)’, namely ‘take 4.5 "hello"’
          In the expression: print $ take 4.5 "hello"

or

    Test.hs:2:21: warning: [-Wdeferred-out-of-scope-variables]
        Variable not in scope: n :: Int

However, those brackets are stripped off by the error sanitizer at https://github.com/google/codeworld/blob/master/codeworld-error-sanitizer/src/ErrorSanitizer.hs#L28 before the message even leaves the server.

I think one could write a rule in ErrorSanitizer.hs to replace "warning: (.\*) \\[-Wdefer-.\*\\]" with "error: \1" on the server.  This would just change the text of the message so the client would render it as an error, but it wouldn't cause the compile to actually fail.  That might have the effect we want.

nixorn: Looks like haskell version of [replace function](https://github.com/google/codeworld/blob/master/codeworld-error-sanitizer/src/RegexShim.hs#L50) does not support group regex replacement.
Can I implement this?

cdsmith: Ah, yes.  I think it would be better to use `subRegex` at http://hackage.haskell.org/package/regex-compat-0.95.1/docs/Text-Regex.html

cdsmith: After another look, it seems that depends on regex-posix, which isn't good.  Among other things, it can't handle unicode characters, which is definitely something we're using now. :(  So, maybe that's just a reference you could use for a new implementation of replacement on top of regex-tdfa?

nixorn: Ok, I happy to try implement this. Anyway in this case we don't need full regex support with nested groups. I'll start from simple replacement without nesting, and will look how it going.

web/js/codeworld_shared.js
cdsmith: This comment is subsumed by the discussion below, so I'll resolve it.  The plan is currently to render with error styling from the beginning, but then append a message when the runtime crash occurs saying: "Giving up because of the error at program.hs:10:42"

web/js/codeworld_shared.js
cdsmith: Sure, whatever works! :)  Just make sure that whatever part you implement is consistent between the two implementations in RegexShim

web/js/codeworld_shared.js
cdsmith: I think all this is still irrelevant.  Again, instead of replacing the old errors, we'd like to just translate the warning to an error at build time, and then replace any deferred type error with the "giving up" message.  I'll resolve this.

nixorn: I almost done this.

nixorn: I about replacing on haskell side. Left one minor issue with correct matching "\[0-9]" pattern to replace group.

nixorn:
My last commit breaks down lot of things. Too early for review.

nixorn:
Okey, this is current version. I changed `replace` function, now it can  do replaces like `warning: ([a-zA-Z -,]*)` to `error: \1`. Take a look please.

codeworld-error-sanitizer/src/RegexShim.hs
```diff
@@ -38,23 +39,45 @@ replace regex replacement str = textFromJSString $
...
 import Data.Monoid
 import qualified Data.Text as T
+import Data.Text.Read (decimal)
 import Text.Regex.Base
 import Text.Regex.TDFA
 import Text.Regex.TDFA.Text
 
+getInserts :: Text -> [(Int, (Int, Int))]
+getInserts replacement = case replacement =~ ("[\\][0-9]" :: Text) :: [MatchArray] of
+    [] -> []
```
cdsmith: You don't need a special case here, because `map` already does the right thing.

codeworld-error-sanitizer/src/RegexShim.hs
```diff
@@ -38,23 +39,45 @@ replace regex replacement str = textFromJSString $
...
+    groups -> fmap processInsertion groups
```
cdsmith: I'd prefer `map` since this is a list.

codeworld-error-sanitizer/src/RegexShim.hs
```diff
@@ -38,23 +39,45 @@ replace regex replacement str = textFromJSString $
...
+    where processInsertion group =
+              case decimal $ T.take len $ T.drop (start + 1) replacement of
```
cdsmith: Did you mean `len - 1` here?  Right now, you're always looking one character beyond the matched expression, which seems confusing.  It sort of works because `decimal` succeeds when it matches only part of the string, but it will do strange things if the character after the group is still a number.

codeworld-error-sanitizer/src/RegexShim.hs
```diff
@@ -38,23 +39,45 @@ replace regex replacement str = textFromJSString $
...
+                  Right (index, _) -> (index, (start, len))
+                  Left _ -> error $ T.unpack $ "Can not parse replacement groups in "
+                                  <> replacement -- should never happen beause handled
+                                                 -- by regex matching above
+              where (start, len) = group ! 0
+
+applyInserts :: Text -> Text -> MatchArray -> [(Int, (Int, Int))] -> Text
```
cdsmith: I don't think this works for more than one group, because you need to fix up the remaining indices in the recursive case.  You might try something like this instead.

    type Inserts = [(Int, (Int, Int))]
    applyInserts :: Text -> Text -> MatchArray -> Inserts -> Text
    applyInserts str replacement matches inserts = go 0 replacement inserts
      where go pos replacement ((index, (rstart, rlen)) : inserts) =
                let pre = T.take (rstart - pos) replacement
                    rest = T.drop (rlen + rstart - pos) replacement
                    (sstart, slen) = matches ! index
                    source = T.take slen $ T.drop sstart str
                in pre <> source <> go (pos + rstart + rlen) rest inserts
            go _ replacement [] = replacement

codeworld-error-sanitizer/src/RegexShim.hs
```diff
@@ -38,23 +39,45 @@ replace regex replacement str = textFromJSString $
...
-replace regex replacement str =
-    let parts = concatMap elems $ (str =~ regex :: [MatchArray])
-    in foldl replaceOne str (reverse parts)
-  where
-    replaceOne :: Text -> (Int, Int) -> Text
-    replaceOne str (start, len) = pre <> replacement <> post
-      where
-        pre = T.take start str
-        post = T.drop (start + len) str
+replace regex replacement str = case str =~ regex of
```
cdsmith: This only matches one occurrence of the target string.  That's different from the current behavior, which matches all occurrences.  I think you still need the fold to handle that.  (This might need more fiddling with indices, though.  Haven't thought it through.)

web/js/codeworld_shared.js
```diff
@@ -1163,6 +1163,10 @@ function printMessage(type, message) {
     if (message.trim() === '') {
         return;
     }
+    // Deferred errors already handled on server side
```
cdsmith: I think you should only handle this case if type === 'error', and you should match a more specific string like "(deferred type error)".  Otherwise, you're likely to ignore a lot of otherwise reasonable messages.

web/js/codeworld_shared.js
```diff
@@ -1163,6 +1163,10 @@ function printMessage(type, message) {
...
+    if (message.indexOf("deferred") != -1){
+        return;
```
cdsmith: Okay, I'm convinced that ignoring the error is good for now.  In the future, I'd like to add some logic where if the program has been running for a while, we add the "Giving up..." message we'd discussed.  But I agree that not including that is the better default, because 99% of the time, the program will crash during the first frame.

codeworld-error-sanitizer/src/RegexShim.hs
nixorn: Yes, you are right.

codeworld-error-sanitizer/src/RegexShim.hs
nixorn: Thanks, fixed.

web/js/codeworld_shared.js
nixorn: Done.

PR #866, opened 2019-03-15
Add layout markers into codeworld-mode

### LINKED ISSUE #54: CodeWorld editor should visually indicate layout
The editor ought to recognize keywords that start layout, and do something to indicate the layout, and help the user understand the significance of lining things up.  Not sure what visual effect is best here.

nixorn:
Closes #54

### DIFF
--- a/web/js/codeworld-mode.js
+++ b/web/js/codeworld-mode.js
@@ -195,11 +195,12 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
                 state.layoutScanState = 'normal';
                 return false;
             } else {
-                state.layoutScanState = 'triggered';
+                state.layoutScanState = 'normal';
                 state.layoutContext.push({
-                    value: 'implicit'
+                    value: 'implicit',
+                    column: column
                 });
-                return false;
+                return true;
             }
         }
 
--- a/web/js/codeworld-mode.js
+++ b/web/js/codeworld-mode.js
@@ -52,6 +52,19 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
     const RE_INCOMMENT = /(?:[^{-]|-(?=$|[^}])|\{(?=$|[^-]))*/;
     const RE_ENDCOMMENT = /-}/;
 
+    const RE_OF = /of/;
+    const RE_LET = /let/;
+    const RE_IN = /in/;
+    const RE_WHERE = /where/;
+    const RE_DO = /do/;
+    const RE_CASE = /case/;
+
+    // Next non-comment or non-whitespace token
+    // is start of layout context
+    const TRIGGERED = 0;
+    // Parsing is not layout sensetive
+    const NORMAL = 1;
+
     function opening(bracket) {
         if (bracket === ')') return '(';
         if (bracket === ']') return '[';
@@ -59,14 +72,36 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
         return bracket;
     }
 
+    // Start new layout context.
+    function checkContextStarter(stream, state) {
+        let match = null;
+        if (stream.match(RE_LET)   ||
+            stream.match(RE_WHERE) ||
+            stream.match(RE_DO)    ||
+            stream.match(RE_CASE)  ||
+            stream.match(RE_OF)
+           ) {
+            let nextChar = stream.peek();
+            // Check if it word like 'lettering' or 'offer'
+            if (nextChar === null || nextChar === undefined || /\s/.exec(nextChar)){
+                state.scanState = TRIGGERED;
+                return true;
+            }
+        }
+        return false;
+    }
+
     // The state has the following properties:
     //
     // func:          The function to tokenize the remaining stream.
     // commentLevel:  Number of levels of block comments.
-    // brackets:      Brackets, from outermost to innermost.
+    // layoutContext: Explicit brackets + implicit layout contexts.
+    // scanState:     When TRIGGERED - next token is start of layout context.
 
     function normal(stream, state) {
-        if (stream.match(RE_WHITESPACE)) return null;
+        let spanStyles = [];
+
+        if (stream.match(RE_WHITESPACE)) return [];
 
         if (stream.match(RE_STARTMETA)) {
             state.func = blockComment('meta');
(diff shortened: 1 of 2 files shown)


cdsmith:
This pull request **introduces 1 alert** when merging 90d81ff360dc8605de1719344405ab1fca15ccf0 into 239eb8d75924e85dbba60588028928a92725fc4a - [view on LGTM.com](https://lgtm.com/projects/g/google/codeworld/rev/pr-52edb5df25c498a940d3e2c90f93f505dcd5d868)

**new alerts:**

* 1 for Unused variable, import, function or class

---

*Comment posted by [LGTM.com](https://lgtm.com)*

web/js/codeworld-mode.js
```diff
@@ -52,21 +52,55 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
+    function checkContextStarter(stream, state) {
+        if (stream.match(RE_LET) ||
+            stream.match(RE_WHERE) ||
+            stream.match(RE_DO) ||
+            stream.match(RE_CASE) ||
+            stream.match(RE_OF)
+        ) {
+            const nextChar = stream.peek();
+            // Check if it word like 'lettering' or 'offer'
+            if (nextChar === null || nextChar === undefined || /\s/.exec(nextChar)) {
```
cdsmith: This looking at the next character seems ugly to me.  Would it work if you added `\b` to the end of the regular expressions above, instead?

web/js/codeworld-mode.js
```diff
@@ -82,43 +116,91 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
-        if (stream.match(RE_VARID) || stream.match(RE_VARSYM)) return 'variable';
-        if (stream.match(RE_CONID) || stream.match(RE_CONSYM)) return 'variable-2';
-        if (stream.match(RE_NUMBER)) return 'number';
-        if (stream.match(RE_CHAR) || stream.match(RE_STRING)) return 'string';
+        if (checkContextStarter(stream, state)) {
+            return ['keyword'];
+        }
+
+        if (state.scanState === NORMAL && stream.match(RE_IN)) {
+            // Cancel previous layout 'let' context
```
cdsmith: You really need to verify the user is in a layout context for `let` here.  For example, imagine the user mistyped: `colored(solidCircle(3), lt c = dark(green) in translucent(c))`  The type of adding `lt` instead of `let` causes `in` to break the parentheses, and it's all chaos.

For that matter, what about this correct code?

    let { x = 5 } in x + 37

Here there is no layout context when you encounter the `in`.  I believe the current code will mistakenly pop some surrounding bracket or layout context.

web/js/codeworld-mode.js
```diff
@@ -82,43 +116,91 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
+            state.layoutContext.pop();
+            return ['keyword'];
+        } else if (state.scanState === NORMAL) {
+            let lastContext = state.layoutContext[state.layoutContext.length - 1];
+            lastContext = lastContext === '{' ? 0 : lastContext;
+            if (stream.sol() && stream.indentation() < lastContext) {
```
cdsmith: I'm not clear on what sol() means here.  Will it ignore whitespace?  Will it ignore comments?

For that matter, why do you need it at all?  If your current column is less than the latest layout context, you must have started a new line, right?

web/js/codeworld-mode.js
```diff
@@ -82,43 +116,91 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
+                // Close previous context
```
cdsmith: I think you need to close layout contexts even if you're in TRIGGERED state.  For instance:

    foo = bar
      where a = case 42 of

    baz = 43

Here, `baz` should still break the layout context for `foo`.

web/js/codeworld-mode.js
```diff
@@ -82,43 +116,91 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
+                state.layoutContext.pop();
+            }
+            if (stream.column() === lastContext || stream.column() === 0) {
```
cdsmith: Why the special case for 0?

web/js/codeworld-mode.js
```diff
@@ -82,43 +116,91 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
+                spanStyles.push('layout');
+            }
+        } else if (state.scanState === TRIGGERED && stream.match('{')) {
+            // Cancel previous layout context because of next char is '{'
+            state.layoutContext.pop();
```
cdsmith: What context are you popping here?  It doesn't look like you pushed the context yet at this point.

web/js/codeworld-mode.js
```diff
@@ -82,43 +116,91 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
+            state.scanState = NORMAL;
+        } else if (state.scanState === TRIGGERED) {
+            state.layoutContext.push(stream.column());
+            spanStyles.push('layout');
+            state.scanState = NORMAL;
```
cdsmith: What about this?

    let
      let x = 4
          y = 6
      z = 7
    in x + y + z

It looks like you're missing a layout context here.  This logic needs to come earlier.

web/js/codeworld-mode.js
```diff
@@ -82,43 +116,91 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
+            return spanStyles;
+        }
 
         if (stream.match(RE_OPENBRACKET)) {
-            state.brackets.push(stream.current());
-            return `bracket${state.brackets.length <= 7 ? `-${  
-                state.brackets.length - 1}` : ''}`;
+            state.layoutContext.push(stream.current());
+            spanStyles.push(
+                `bracket${state.layoutContext.length <= 7 ? `-${
```
cdsmith: I think the right thing to do here is count the explicit brackets in the layout context.  You should not bump the bracket count when there's an implicit layout.

web/js/codeworld-mode.js
```diff
@@ -82,43 +116,91 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
-                return 'bracket';
+                spanStyles.push('bracket');
+                return spanStyles;
             } else {
-                while (state.brackets.length > i) state.brackets.pop();
-                return `bracket${state.brackets.length <= 6 ? `-${ 
-                    state.brackets.length}` : ''}`;
+                while (state.layoutContext.length > i) state.layoutContext.pop();
+                spanStyles.push(
+                    `bracket${state.layoutContext.length <= 6 ? `-${
```
cdsmith: Same comment about counting explicit brackets, not all contexts.

web/js/codeworld-mode.js
```diff
@@ -170,19 +252,21 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
         },
 
         token: (stream, state) => {
             const t = state.func(stream, state);
-            if (['variable', 'variable-2'].indexOf(t) !== -1) {
-                const w = stream.current();
-                if (wellKnownWords.hasOwnProperty(w)) return wellKnownWords[w];
+            const w = stream.current();
+            if (wellKnownWords.hasOwnProperty(w)) {
+                if (t.indexOf('variable') !== -1) t.pop(t.indexOf('variable'));
```
cdsmith: I don't think Array.pop takes an argument.  I think you can do `t.splice(t.indexOf(...), 1)`  (note that I'm using `splice`, not `slice` which looks similar, but does something different)

web/js/codeworld-mode.js
nixorn: TRIGGERED does mark next token as open of new context, wherever it is.
image

For your example there should be condition like "if current column less than `where` indentation - do not open new context". Or even close all contexts, which context starter indentation more than current column. I see next solution - make context entries `(contextStarterColumn, contextColumn)` instead of just `contextColumn` integer value.

web/js/codeworld-mode.js
nixorn: Now I see it is consequence of error with context closing. `state.layoutContext` is inconsistent now, as you said above it does not close contexts correctly, so `stream.column() === 0` is wrong way to fix

image

I'll fix context closing and delete this.

web/js/codeworld-mode.js
nixorn: Yes, it is error.

web/js/codeworld-mode.js
nixorn: Thanks!

web/js/codeworld-mode.js
nixorn: Yes, it is leftover I forgot to delete.

web/js/codeworld-mode.js
nixorn: Yes, that's much better.

web/js/codeworld-mode.js
```diff
@@ -52,21 +52,55 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
         if (bracket === '}') return '{';
         return bracket;
     }
 
+    // Start new layout context.
+    function checkContextStarter(stream, state) {
+        if (stream.match(RE_LET) ||
+            stream.match(RE_WHERE) ||
+            stream.match(RE_DO) ||
+            stream.match(RE_CASE) ||
```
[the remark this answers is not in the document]
nixorn: Got it.

web/js/codeworld-mode.js
nixorn: Okey, this is about upgrading the `layoutContext` to store more information.

Now `layoutContext` contain brackets and integers. Thinking about transform it to something like
```
[{type: 'bracket', contextStarterColumn: 0}
,{type: 'square', contextStarterColumn: 1}
,{type: 'curly', contextStarterColumn: 2}
,{type: 'where',  contextStarterColumn: 8, contextColumn: 14}
,{type: 'let', contextStarterColumn: 14, contextColumn: 18}]
```

web/js/codeworld-mode.js
nixorn: I not got this example, sorry
image

web/js/codeworld-mode.js
cdsmith: This looks like a good plan.  I'm not sure I understand what `contextStarterColumn` is about, but the rest seems right.

web/js/codeworld-mode.js
cdsmith: My mistake.  I gave you a bad example.  Another one might be:

    do
        do one
           two
        three

The second `do` starts a new layout context, but also triggers nested layout.  On further though, I think this is only possible with `do`, and granted I don't really use `do` in codeworld-mode.  It's still possible, though, and not hard to fix.

cdsmith:
@nixorn: Are you blocked on anything, or still working on updates?

nixorn:
@cdsmith still working.

cdsmith:
This pull request **introduces 1 alert** when merging 0ad83b136f818a35fb58b40ddc314b429e478f4a into 239eb8d75924e85dbba60588028928a92725fc4a - [view on LGTM.com](https://lgtm.com/projects/g/google/codeworld/rev/pr-fe6c66529adf7ddd2c153b624b2771d21df880e0)

**new alerts:**

* 1 for Unused variable, import, function or class

---

*Comment posted by [LGTM.com](https://lgtm.com)*

web/js/codeworld-mode.js
```diff
@@ -47,26 +47,231 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
+
     function opening(bracket) {
         if (bracket === ')') return '(';
         if (bracket === ']') return '[';
         if (bracket === '}') return '{';
         return bracket;
     }
 
+    // Check if there start of new layout context.
+    // Returns {style: <style of current token>, context:<layout context>}
```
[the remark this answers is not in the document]
nixorn: This advice looks extremely helpful.

nixorn:
I see incorrect handling of 'start' state. Fixing.

nixorn:
Done.

PR #867, opened 2019-03-21
feat: Requirements Checker UI only appears when no runtime errors are…

### LINKED ISSUE #865: Deferred errors makes requirements checker less useful
**Shared Program**: https://code.world/#PYurNtPA2Z7wBP-5D1xq0rA

**Description**

It used to be that the requirements checker didn't run unless your program worked.  This was because of https://github.com/google/codeworld/commit/b4e36e443789ae2bf193d2c36de3e369fb295b0e and was done deliberately because it was confusing to students that the requirements checker marks their project as correct even though the code doesn't compile.

Unfortunately, now that we defer many errors until runtime, we're back in a state where the requirements checker succeeds when there are build errors.  I see a few possibilities here.

1. Add some logic to detect deferred errors at build time and suppress the requirement checker.  However, this has the disadvantage that if the deferred error is truly unnecessary (e.g., in an unreferenced variable), the requirements checker doesn't appear to run even when the program appears to work.

2. Extend the requirements language (or just add a silent requirement by default) to require that there are no errors, and display an actual requirements dialog, but just visibly fail because there are errors.

3. Push forward to doing runtime verification, too, and then the requirement checker would fail during runtime verification.  But this is a very speculative project, and it's a long way off.

4. Show the requirement checker UI only after the program doesn't crash for a couple seconds.  Kludgy, but it doesn't sound too hard.

5. Just disable deferred errors if there are requirements.  I'd prefer not to...

Currently, I favor option 4 for the solution.

devanshbatra04:
closes #865

### DIFF
--- a/web/js/codeworld.js
+++ b/web/js/codeworld.js
@@ -804,6 +804,8 @@ function stop() {
 }
 
 function run(hash, dhash, msg, error, generation) {
+    if (!msg && !error) return;
+
     window.runningGeneration = generation;
 
     const runner = document.getElementById('runner');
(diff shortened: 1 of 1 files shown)


web/js/codeworld.js
```diff
@@ -804,6 +804,8 @@ function stop() {
 }
 
 function run(hash, dhash, msg, error, generation) {
+    if (!msg && !error) return;
```
cdsmith: What's this for?  It looks like it might result in forgetting to run code that doesn't have warnings?

cdsmith:
Thanks.  This looks right, but I have one question (posted above).

web/js/codeworld.js
devanshbatra04: I saw the function ```run``` being executed with these trivial arguments when the page loads up with source code already present. I saw the code was essentially doing nothing, so I placed the return statement at top. 

Should I undo this? Everything works just as well otherwise.
Screenshot from 2019-03-22 12-52-41

cdsmith: Yes, I think what you're seeing is a call to the `stop` function.  This may not be necessary on page load, but it's definitely important in other places. I'll merge now and then revert this.

devanshbatra04: Great. Thank you!

PR #868, opened 2019-03-22
WIP (Add tooltip for inline functions) 

### LINKED ISSUE #850: Tool tip with argument help when typing a function
When typing a built-in function, it would be neat if a tooltip popped up above the line, with names for the parameters, and a bold font to show which one you're currently on.

bcpcv0s34pw

devanshbatra04:
closes #850 

The gif give a brief idea abut the progress. I dont understand where am I going wrong with the CSS? The tooltip should be perfectly solid, I copied the styles from ```CodeMirror-hover-tooltip``` class.

ezgif com-video-to-gif

### DIFF
--- a/web/css/codeworld-cm.css
+++ b/web/css/codeworld-cm.css
@@ -164,18 +164,6 @@ div.CodeMirror span.cm-matchhighlight {
 
 /* Style overrides for the autocomplete box and doc-on-hover. */
 
-.function-tooltip-styling {
-    background: #ffe;
-    border-radius: 0.5em;
-    border: 1px solid #888;
-    box-shadow: 0.1em 0.1em 0.1em gray;
-    min-height: 1em;
-    max-height: 20em;
-    min-width: 5em;
-    overflow: hidden;
-    padding: 0.3em 0.5em;
-}
-
 #function-details {
     z-index: 10;
 }
@@ -216,7 +204,8 @@ li.CodeMirror-hint-active {
 
 .CodeMirror-hover-tooltip,
 .CodeMirror-lint-tooltip,
-.hint-description {
+.hint-description,
+.function-tooltip-styling {
     background-color: #ffe;
     border-radius: 0.5em;
     border: 1px solid #888;
--- a/web/css/codeworld-cm.css
+++ b/web/css/codeworld-cm.css
@@ -164,6 +164,22 @@ div.CodeMirror span.cm-matchhighlight {
 
 /* Style overrides for the autocomplete box and doc-on-hover. */
 
+.function-tooltip-styling {
+    background: #ffe;
+    border-radius: 0.5em;
+    border: 1px solid #888;
+    box-shadow: 0.1em 0.1em 0.1em gray;
+    min-height: 1em;
+    max-height: 20em;
+    min-width: 5em;
+    overflow: hidden;
+    padding: 0.3em 0.5em;
+}
+
+#function-details {
+    z-index: 10;
+}
+
 .CodeMirror-hint,
 .hover-decl {
     color: #555;
--- a/web/js/codeworld-mode.js
+++ b/web/js/codeworld-mode.js
@@ -28,7 +28,6 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
         '\\\\HT|\\\\LF|\\\\VT|\\\\FF|\\\\CR|\\\\SO|\\\\SI|\\\\DLE|\\\\DC1|\\\\DC2|' +
         '\\\\DC3|\\\\DC4|\\\\NAK|\\\\SYN|\\\\ETB|\\\\CAN|\\\\EM|\\\\SUB|\\\\ESC|' +
         '\\\\FS|\\\\GS|\\\\RS|\\\\US|\\\\SP|\\\\DEL';
-
     const RE_WHITESPACE = /[ \v\t\f]+/;
     const RE_STARTMETA = /{-#/;
     const RE_STARTCOMMENT = /{-/;
@@ -86,18 +85,34 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
         }
 
         if (stream.match(RE_QUAL)) return 'qualifier';
-        if (stream.match(RE_VARID) || stream.match(RE_VARSYM)) return 'variable';
-        if (stream.match(RE_CONID) || stream.match(RE_CONSYM)) return 'variable-2';
+        let varType = null;
+        if (stream.match(RE_VARID) || stream.match(RE_VARSYM)) varType = 'variable';
+        if (stream.match(RE_CONID) || stream.match(RE_CONSYM)) varType = 'variable-2';
+        if (varType) {
+            if (stream.string[stream.pos] === '(') {
+                state.encFunc.push({
+                    functionName: stream.current(),
+                    roundBracketLevel: state.roundBracketLevel,
+                    argIndex: 0
+                });
+            }
+            return varType;
+        }
         if (stream.match(RE_NUMBER)) return 'number';
         if (stream.match(RE_CHAR) || stream.match(RE_STRING)) return 'string';
 
         if (stream.match(RE_OPENBRACKET)) {
+            if (stream.current() === '(') state.roundBracketLevel++;
             state.brackets.push(stream.current());
             return `bracket${state.brackets.length <= 7 ? `-${  
                 state.brackets.length - 1}` : ''}`;
         }
 
         if (stream.match(RE_CLOSEBRACKET)) {
+            if (stream.current() === ')') {
+                state.roundBracketLevel--;
+                if (state.encFunc.length && state.encFunc[state.encFunc.length - 1].roundBracketLevel === state.roundBracketLevel) state.encFunc.pop();
+            }
             const i = state.brackets.lastIndexOf(opening(stream.current()));
             if (i < 0) {
                 return 'bracket';
@@ -108,7 +123,9 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
             }
         }
 
-        if (stream.eat(',')) return null;
+        if (stream.match(',')) {
+            if (state.encFunc.length) state.encFunc[state.encFunc.length - 1].argIndex++;
+        }
 
         stream.next();
         return 'error';
@@ -170,7 +187,9 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
             return {
                 func: normal,
                 commentLevel: 0,
-                brackets: []
+                brackets: [],
+                encFunc: [],
+                roundBracketLevel: 0
             };
         },
 
--- a/web/js/codeworld-mode.js
+++ b/web/js/codeworld-mode.js
@@ -88,35 +88,12 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
         }
 
         if (stream.match(RE_QUAL)) return 'qualifier';
-        let varType = null;
-        if (stream.match(RE_VARID) || stream.match(RE_VARSYM)) varType = 'variable';
-        if (stream.match(RE_CONID) || stream.match(RE_CONSYM)) varType = 'variable-2';
-        if (varType) {
-            if (stream.string[stream.pos] === '(') {
-                state.encFunc.push({
-                    functionName: stream.current(),
-                    roundBracketLevel: state.roundBracketLevel,
-                    argIndex: 0
-                });
-            }
-            return varType;
-        }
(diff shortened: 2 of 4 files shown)


cdsmith:
Wow, that's pretty far along already.  There are a few places where we do weird things with z-index in the CSS, and it looks like that might be tripping you up.  For example: https://github.com/google/codeworld/blob/3a750230b8e0ab9ccdf46b2e5f4cac7956a6217c/web/css/codeworld-cm.css#L230

devanshbatra04:
Thanks, I will try something out. :)

devanshbatra04:
ezgif com-video-to-gif

This is the output right now. Slightly stuck on efficiently highlighting the right argument. Will make the push in a few hours. 
One Question: Is there an already implemented way to get the argument names? Right now I only get the types. If not I will get cracking.

devanshbatra04:
Screenshot from 2019-03-31 07-29-32

web/css/codeworld-cm.css
```diff
@@ -164,6 +164,22 @@ div.CodeMirror span.cm-matchhighlight {
 
 /* Style overrides for the autocomplete box and doc-on-hover. */
 
+.function-tooltip-styling {
```
cdsmith: Rather than copying this, can you split up the existing rule below, and add this new class to the list for the first part?

web/js/codeworld.js
```diff
@@ -221,6 +221,59 @@ function initCodeworld() {
...
+        docDiv.appendChild(annotation);
+
+        topDiv.appendChild(docDiv);
+        window.codeworldEditor.addWidget(cursor, topDiv, true, "above", "left");
+    });
+
+    window.getEnclosingFunction = (cm, cur = cm.getCursor()) => {
+        const token = cm.getTokenAt(cur);
+        const line = cm.getLine(cur.line);
+        if (! line.includes('(')) return;
```
cdsmith: I think you meant `line.substr(0, token.end).includes(...)` here?  Otherwise you might continue, but later fail to find the parenthesis.

web/js/codeworld.js
```diff
@@ -221,6 +221,59 @@ function initCodeworld() {
...
+        const annotation = document.createElement('div');
+        renderDeclaration(annotation, enclosingFunction.string, keywordData, 9999, argIndex);
+        annotation.className = 'hover-decl';
+        docDiv.appendChild(annotation);
+
+        topDiv.appendChild(docDiv);
+        window.codeworldEditor.addWidget(cursor, topDiv, true, "above", "left");
+    });
+
+    window.getEnclosingFunction = (cm, cur = cm.getCursor()) => {
```
cdsmith: I'm pretty sure this is insufficient.  For example, consider a cursor right after `sin(x) + `.  Won't this tell you that `sin` is the enclosing function?  You actually need to scan backwards and match brackets until the last unmatched open bracket.

Another option (and my first instinct is that it would be better) would be to compute this during lexing in codeworld-mode.js, and store it in the token state.  Then you don't have to do so much rescanning of the text when the cursor moves.  CodeMirror would take care of all the logic needed to keep that state up-to-date as the document changes.  And your interpretation of the source code would be more consistent with syntax highlighting -- for example, when something like layout rules force the editor to conclude that there's an unmatched bracket, this code would be consistent with that assumption.  You'd also get the feature to work for line-wrapped functions, for free.

web/js/codeworld.js
```diff
@@ -221,6 +221,59 @@ function initCodeworld() {
...
+                    getEnclosingFunction(
+                        window.codeworldEditor,
+                        {
+                            line: cursor.line, ch: elem.end
+                        })
+                        .string === enclosingFunction.string
+                    && elem.type && elem.type !== "keyword"
+                    && elem.type!=="bracket-0"
+                    && elem.type !== "built-in"
+            ).findIndex(elem => (elem.start === currentToken.start || elem.end === currentToken.start || elem.start === currentToken.end || elem.end === currentToken.end));
```
cdsmith: Are you sure that counting tokens works here?  There are lots of things, like passing an expression as an argument, that  would change the number of tokens.  If the arguments are in a tuple, I'd expect you to look for commas in particular.

web/js/codeworld.js
```diff
@@ -221,6 +221,59 @@ function initCodeworld() {
...
+            lineTokens.filter(
+                elem =>
+                    getEnclosingFunction(
+                        window.codeworldEditor,
+                        {
+                            line: cursor.line, ch: elem.end
+                        })
+                        .string === enclosingFunction.string
+                    && elem.type && elem.type !== "keyword"
+                    && elem.type!=="bracket-0"
```
cdsmith: "bracket-0" is almost surely wrong.  Did you want to look for brackets at the same level as the open-bracket that introduced the function?  (But see the other comment about moving this logic to the editor mode instead; this reinforces that suggestion.)

web/js/codeworld.js
devanshbatra04: Got it. I think I can solve all of the problems by doing something like
```javascript
line = line.substr(0, token.end);
let level = 0;
let i = line.length - 1;
let argIndex = 0;
for (; i>=0; i--){
if (line[i] === ')') level++;
if (line[i] === '(' && level === 0) break;  
if (line[i] === '(' && level > 0) level--;  
if (level > 0) continue;
if (line[i] === ',') argIndex++;  // (no. of commas in current level = argIndex)
}
const enclosing function = // logic to get token preceding index and see if it's declaration is known.
```

web/js/codeworld.js
devanshbatra04: looking over this method, if it doesn't work out I guess I will go with the method I mentioned above.

web/js/codeworld.js
devanshbatra04: This isn't required after implementing your approach.

web/js/codeworld.js
devanshbatra04: @cdsmith I have implemented your approach and it works like a charm. Please have a look :)

devanshbatra04:
I had been working on an outdated pull apparently. And kind of re implemented @nixorn's context feature. How embrassing is that. I will fix that asap.

web/js/codeworld-mode.js
```diff
@@ -89,12 +88,35 @@ CodeMirror.defineMode('codeworld', (_config, modeConfig) => {
...
 
         if (stream.match(RE_QUAL)) return 'qualifier';
-        if (stream.match(RE_VARID) || stream.match(RE_VARSYM)) return 'variable';
-        if (stream.match(RE_CONID) || stream.match(RE_CONSYM)) return 'variable-2';
+        let varType = null;
+        if (stream.match(RE_VARID) || stream.match(RE_VARSYM)) varType = 'variable';
+        if (stream.match(RE_CONID) || stream.match(RE_CONSYM)) varType = 'variable-2';
+        if (varType) {
+            if (stream.string[stream.pos] === '(') {
+                state.encFunc.push({
```
cdsmith: This might be less intrusive inside the context array.  If each context in the stack had attributes for `functionName` and `argIndex`, you could modify the object at the top of the stack.  The right place to do that is probably in `updateLayout` (below), instead of during tokenization.

You shouldn't need to keep track of the parenthesis nesting level (`roundBracketLevel`) at all.  Those function call contexts should already be popped from the stack at the right time by the existing code.  As for incrementing the index when you see a comma, it's actually a bug that commas don't break layout now.  (The Note 5 layout rule implies that they should.)  After fixing that, the top context on the stack should always be the right one to increment `argIndex`.

web/js/codeworld_shared.js
```diff
@@ -234,15 +234,38 @@ function renderDeclaration(decl, keyword, keywordData, maxLen) {
...
         if (keywordData.symbolEnd + leftover.length > maxLen && leftover.length >
             3) {
             leftover = `${leftover.slice(0, maxLen - 3 - keywordData.symbolEnd) 
             }...`;
         }
-        decl.appendChild(document.createTextNode(leftover));
+    }
+    if (argIndex >= 0) {
+        if (!leftover.length) {
+            leftover = " :: Could not decipher arguments";
```
cdsmith: I'm assuming this message is there for debugging?  In production, it would be better to not show the type than to complain about not being able to figure it out.

devanshbatra04: Okay. i will remove this.

web/js/codeworld-mode.js
devanshbatra04: I am having a tough time trying to understand contexts. Any issue or documentation that might be helpful? The code comments weren't very clear.

As far as I can tell contexts do not (currently) list the line number, only the column number. Could this be a problem later if the parenthesis are spread across multiple lines?

cdsmith: There's not really documentation.  Contexts are just grouping levels within the code.  This grouping can be explicit brackets of any of the three main types (parenthese, square brackets, or curly braces), or they can be the implicit curly braces added by the Haskell layout algorithm.

If you're not familiar with layout, basically, in Haskell certain keywords start an indentation-sensitive grouping if they aren't followed by curly braces.  e.g.,

    f x = a + b
      where a = x^2
            b = x^3

The `where` keyword triggers layout, so the compiler remembers the column number for the next character (except for whitespace or comments) after the where block, and adds braces and semicolons around the lines starting in that same column.  So this example means the same thing as:

    f x = a + b where { a = x^2 ; b = x^3 }

In the editor mode, a context is an object representing that grouping (whether it's explicitly typed or just implied), and its `value` property tells you what kind of grouping it is: `(`, `[`, `{`, or if it's an implicit context, the keyword that introduced it (`where`, `let`, `of`, `do`, or `\ case`).

Don't worry about the column number.  You shouldn't need *either* the line or column, actually.  The only reason the column number is there is that the layout algorithm depends on column numbers: new lines starting in the same column are separated by an implied semicolon, and the next line starting in an earlier column ends the context.  But that's just implementation detail.  Ignore it.

All you need to do is: (a) when you see an open parenthesis, check the last token (in state.lastTokens), and if it's a variable and not a keyword, add it to the new layout context as `functionName`, and add an `argIndex` of 0, then (b) when you see a comma, start by popping any implicit contexts off the end of the stack, and then if the top context has an `argIndex`, increment it, and then (c) when you want to trigger your popup, look for the top context that has a `functionName` in it, and use that.

devanshbatra04: This definitely helped a lot. I really appreciate you giving valuable time in making things clear. Thank you very much for all the detail. :heart:

devanshbatra04: made the required changes. Hope it works out.

cdsmith:
This looks good to me.  I haven't done any testing, so I'm just going by the code and the gifs you've posted.  Are there any concerns, or questions you have before merging?

cdsmith:
(Yeah, I know there are no argument names, but this seems worth merging in its current form, and the help text can be changed later.)

devanshbatra04:
No questions. I will make another PR for the argument names thing. 
I think we will need to change ```codeworld-base.txt``` to get argument names for all builtin functions. Those in the current scope seem easy to do in ```parseSymbolsFromCurrentCode``` in codeworld_shared.

web/js/codeworld_shared.js
```diff
@@ -245,20 +245,14 @@ function renderDeclaration(decl, keyword, keywordData, maxLen, argIndex = -1) {
...
-                leftover.lastIndexOf("(") + 1,
-                leftover.lastIndexOf(")")
-            ).split(",").map(token => token.trim());
-            const ReturnType  = leftover.split("->")[1].trim();
-            argIndex = Math.min(argIndex, tokens.length - 1);
-            tokens[argIndex] = `<strong>${tokens[argIndex]}</strong>`;
-            leftover = ` :: (${tokens.join(', ')}) -> ${ReturnType}`;
-
-        }
+        let tokens = leftover.substring(
```
cdsmith: I'm not sure what this is going to do if the declaration looks different from the way you expect.  Another (safer) approach might be to match a regular expression to pull out the arguments, like:

    const [start, args, remainder] = /^(\s*::\s*[(])(.*)([)]\s*->.*)$/.exec(leftover)

You'll also want another pattern for functions of one argument, since there are no parentheses in their types.  (I suspect these don't work in the current code?)

devanshbatra04: Understood. I will work this out in a few hours.

cdsmith: Great.  I'm going to bed now, but I'll merge this tomorrow when I have time to do some manual testing.

devanshbatra04: I came up with a regex that works for both single arg and multiple arg declarations 
```javascript
const [head, args, tail] = (/^(\s*::\s*[(]*)([\w,\s]*)([)]*\s*->.*)$/).exec(leftover).slice(1)
```

PR #887, opened 2019-03-29
WIP Fix broken time-step ellision

### LINKED ISSUE #681: The time-step elision optimization is broken
We broke time-step ellision for interactions somewhere.  See https://code.world/haskell#PPUTRWUDaX3tF9KMQYGTS-w

nixorn:
Closes #681

### DIFF
--- a/codeworld-api/src/CodeWorld/CanvasM.hs
+++ b/codeworld-api/src/CodeWorld/CanvasM.hs
@@ -85,6 +85,56 @@ saveRestore m = do
 
 #ifdef ghcjs_HOST_OS
 
+newtype StubCanvasM a = StubCanvasM
+    { unStubCanvasM :: IO a
+    } deriving (Functor)
+
+
+instance Applicative StubCanvasM where
+    pure x = StubCanvasM (return x)
+    f <*> x = StubCanvasM (unStubCanvasM f <*> unStubCanvasM x)
+
+instance Monad StubCanvasM where
+    return = pure
+    m >>= f = StubCanvasM (unStubCanvasM m >>= unStubCanvasM . f)
+
+instance MonadIO StubCanvasM where
+    liftIO = StubCanvasM
+
+instance MonadCanvas StubCanvasM where
+    type Image StubCanvasM = ()
+    save = StubCanvasM $ return ()
+    restore = StubCanvasM $ return ()
+    transform _ _ _ _ _ _ = StubCanvasM $ return ()
+    translate _ _ = StubCanvasM $ return ()
+    scale _ _ = StubCanvasM $ return ()
+    newImage _ _ m = StubCanvasM $ do
+        a <- unStubCanvasM m
+        return ((), a)
+    drawImage _ _ _ _ _ = StubCanvasM $ return ()
+    globalCompositeOperation _ = StubCanvasM $ return ()
+    lineWidth _ = StubCanvasM $ return ()
+    strokeColor _ _ _ _ = StubCanvasM $ return ()
+    fillColor _ _ _ _ = StubCanvasM $ return ()
+    font _ = StubCanvasM $ return ()
+    textCenter = StubCanvasM $ return ()
+    textMiddle = StubCanvasM $ return ()
+    beginPath = StubCanvasM $ return ()
+    closePath = StubCanvasM $ return ()
+    moveTo _ = StubCanvasM $ return ()
+    lineTo _ = StubCanvasM $ return ()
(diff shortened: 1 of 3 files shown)


codeworld-api/src/CodeWorld/CanvasM.hs
```diff
@@ -85,6 +85,56 @@ saveRestore m = do
...
+instance Monad StubCanvasM where
+    return = pure
+    m >>= f = StubCanvasM (unStubCanvasM m >>= unStubCanvasM . f)
+
+instance MonadIO StubCanvasM where
+    liftIO = StubCanvasM
+
+instance MonadCanvas StubCanvasM where
+    type Image StubCanvasM = ()
+    save = StubCanvasM $ return ()
```
cdsmith: I suppose since you have a Monad instance already, there's no need to explicitly wrap in StubCanvasM.  You could just write `save = return ()`.  Same for others.

nixorn:
What will be next steps? I guess I should write some haskell code which going to be compiled into js via ghcjs and then evaluated in nodejs as test?

nixorn:
Thinking about how I could patch `Driver.hs` to use with ghcjs.
I think need
1. get some code like https://code.world/haskell#PPUTRWUDaX3tF9KMQYGTS-w
2. write different `Driver.hs` primitives, which allow success to state per step besides execution. something like `runNStepsTest :: Int -> world -> (Event -> World -> World) -> IO [World]`
3. make function which collect results for test. `collect :: [World] -> Either SuccessMsg ErrorMsg`
4. compile this with ghcjs
5. run with nodejs as HUnit tests.

cdsmith:
Just wanted to point out that you can run unit tests with GHCJS without that wrapper.  The unit test should be the client code, and you can do this:

    cd codeworld-api
    cabal configure --ghcjs --enable-tests
    cabal test

You'll want to write your tests directly against codeworld-api (that is, in the style of http://code.world/haskell), but that should be fine.

cdsmith:
I just merged a couple commits to get this started. It's a lot harder than I thought it would be.  https://github.com/google/codeworld/commit/193b9aef09a8db59b923105ba377fd1a3ffc1e88 is a good pattern for the unit test, and https://github.com/google/codeworld/commit/700645a19b1b1ed75eb99d03f84f59eba7190dd4 gets some of the way toward abstracting over the underlying graphics environment.  The latter change was way harder than I though because unfortunately `blank-canvas` lacks a built-in MonadIO instance, so I had to hack around that to get a functioning one.

nixorn:
From your email
"Unfortunately, I didn't get far enough to take advantage of this work by simplifying the CodeWorld.Driver module. There's still a lot to be done before that's possible.  We need some abstraction over how we obtain the original canvas, and how we deliver events."

1. What means "original" canvas? For browser obtaining canvas is `getElementById`. For nodejs it should be like some dummy element which swallow all commands to it. Should I use conditions with `#ifdef` as described here https://guide.aelve.com/haskell/cpp-vww0qd72 to tell in which environment execution are? 

2. What is delivering of event? In snippet from original issue generated js application spawns `TimePassing` events from browser. Also `KeyPress`, etc. Event handler describes how to react to event, but event generated by "user" as I see here https://github.com/google/codeworld/blob/master/codeworld-api/src/CodeWorld/Event.hs#L58 . Even `TimePassing` is generating by "user". Is " some abstraction over ... how we deliver events" is about possibility to mock events with stub canvas in tests? E.g. generate fake `KeyPress` event in tests?

cdsmith:
Yes, in both cases, I'm referring to the top-level code that handles the event loop, listeners, etc.  This is the last bit of code that needs work so that it can be tested with a stubbed graphics implementation.  Unfortunately, this is where the logic you need to test lives.

It's difficult to explain precisely what needs to be done here without just doing it.  I'm figuring it out myself as I go.  So if you think you understand how to make this work, go for it.  Otherwise, I cam happy to just merge the StubCanvasM implementation you have, and work on the unit test myself.  I don't think I'm too far off, really.  Once there's a failing test, it could be much easier for you to dig into why it's failing and work on a solution.

If you think you can make progress on your own, though, then go for it!

nixorn:
Ok :)

nixorn:
I meant I want to continue fix this. Issue not resolved.

cdsmith:
Sorry, I didn't mean to close the issue.  Just to merge the stub so that I can work on building a failing unit test to demonstrate the issue.

cdsmith:
Issue is reopened.

PR #893, opened 2019-04-03
function tooltip: handled exception and render near function call

devanshbatra04:
Screenshot from 2019-04-04 02-44-01


web/js/codeworld.js
```diff
@@ -246,12 +246,23 @@ function initCodeworld() {
...
+            topDiv.remove();
+            return;
+        }
         annotation.className = 'hover-decl';
         docDiv.appendChild(annotation);
 
         topDiv.appendChild(docDiv);
-        window.codeworldEditor.addWidget(cursor, topDiv, true, "above", "left");
+        const lineTokens = window.codeworldEditor.getLineTokens(cursor.line);
+        const functionToken = lineTokens.filter(tkn => tkn.end <= currentToken.start && tkn.string === functionName).pop();
```
cdsmith: This might be the wrong token.  For instance, if I write `rotated(rotated(solidRectangle(1,1), 45, 30)`,  the `30` token will fine the wrong `rotated` as a function name.

On the other hand, this might not find the token at all.  If I write:

    rotated(
        solidRectangle(1,1),
        45)

The function name is on a previous line.  In this case, I suspect your code with throw an exception.

Another approach might be to change the scanner again, to keep more information (e.g., line and column number of the function name) in the context.  Then you'll have it available here for use.

devanshbatra04: Interesting, I will do that. Thanks. :-)

devanshbatra04: I did that but found out that we didn't need the function column number and actually rendering the tooltip at the same place as the function had some issues. It made sense to have it in the same line, so didn't need the line number as well. 

Besides I think it is more important that the error/exception gets handled in the master.

cdsmith:
Did you forget to add codeworld-mode.js to your commit?

devanshbatra04:
Isnt needed. ```Col - function.length``` does the trick

devanshbatra04:
I see there are no contexts yet in the Codemirror mode used in [code.world/haskell](https://code.world/haskell) and that is emitting console errors in almost all programs on moving the cursor. I will correct that issue in the same PR.
Check out https://code.world/haskell#PvJdMFyZGH1k6eqnWQBKK5A

Sorry for all the mess
Screenshot from 2019-04-05 01-35-09

cdsmith:
Oh, yes.  The contexts field is written by codeworld-mode.js, which only runs for codeworld-mode.  Haskell mode uses CodeMirror's standard "haskell" mode, so it won't work.  A fix would be great.

PR #894, opened 2019-04-04
Tests to reproduce 681

nixorn:


nixorn:
By the way, what do you think about headless chrome instead of nodejs for testing?

cdsmith:
I think nodejs is already installed and run by default in GHCJS unit tests.  For non-graphics issues, that seems like the right way to go.  If, in the future, we want to test the accuracy of graphics code, it might be worth looking into a more complete browser implementation.

nixorn:
We need to adapt our codebase to work in different environment.
So we anyway need to adapt codebase, possibility to throw events or access to state during execution is required, but if is possible easy run tests using browser environment - no need to make things like StubCanvas.

More system dependencies for testing, but less `#ifdef-s`...

cdsmith:
Sure, but I don't think running in headless chrome fixes this problem anyway.  You want to capture how often the step function is iterated, and this is just as invisible to headless chrome as it is to node.js.  Drawing code is not relevant here, and stubbing it out is the right thing to do.

nixorn:
Ok.

nixorn:
I want to try rename `StubCanvasM` back to `CanvasM`. We already control which version of `CanvasM` is used by preprocessor commands.

nixorn:
```bash
$ cabal test
Preprocessing library for codeworld-api-0.3.1..
Building library for codeworld-api-0.3.1..
Preprocessing test suite 'unit-tests' for codeworld-api-0.3.1..
Building test suite 'unit-tests' for codeworld-api-0.3.1..
Linking dist/build/unit-tests/unit-tests.jsexe (Main,TestDriver)
Running 1 test suites...
Test suite unit-tests: RUNNING...
Test suite unit-tests: PASS
Test suite logged to: dist/test/codeworld-api-0.3.1-unit-tests.log
1 of 1 test suites (1 of 1 test cases) passed.
```
but
```bash
$ cat dist/test/codeworld-api-0.3.1-unit-tests.log
Test suite unit-tests: RUNNING...
### Failure in: 0:Time step elision
test/TestDriver.hs:15
Is time step ellision ok
Cases: 1  Tried: 1  Errors: 0  Failures: 1
Test suite unit-tests: PASS
Test suite logged to: dist/test/codeworld-api-0.3.1-unit-tests.log
```

What is reason for this?

cdsmith:
I'm looking at this now.  `main = exitFailure` as a test definitely succeeds, so it's not ignoring the exit code.  I have no idea what you're seeing.

cdsmith:
Okay, I don't see the same thing:

    $ ./test.sh 
    ===== IN codeworld-api, RUNNING: cabal_install --ghcjs --enable-tests --only-dependencies
    Warning: the --global flag is deprecated -- it is generally considered a bad
    idea to install packages into the global store
    Resolving dependencies...
    All the requested packages are already installed:
    Use --reinstall if you want to reinstall anyway.
    ===== IN codeworld-api, RUNNING: cabal_configure --ghcjs --enable-tests
    Warning: the --global flag is deprecated -- it is generally considered a bad
    idea to install packages into the global store
    Resolving dependencies...
    Warning: The following packages are likely to be broken by the reinstalls:
    codeworld-base-0.2.0.0
    Continuing even though the plan contains dangerous reinstalls.
    Configuring codeworld-api-0.3.1...
    Building codeworld-api-0.3.1...
    Installed codeworld-api-0.3.1
    ===== IN codeworld-api, RUNNING: cabal test
    Preprocessing library for codeworld-api-0.3.1..
    Building library for codeworld-api-0.3.1..
    Preprocessing test suite 'unit-tests' for codeworld-api-0.3.1..
    Building test suite 'unit-tests' for codeworld-api-0.3.1..
    Linking dist/build/unit-tests/unit-tests.jsexe (Driver,Main)
    Running 1 test suites...
    Test suite unit-tests: RUNNING...
    Driver:
      time step ellision: [Failed]
    expected: 0
     but got: 1

             Test Cases  Total      
     Passed  0           0          
     Failed  1           1          
     Total   1           1          
    Test suite unit-tests: FAIL
    Test suite logged to: dist/test/codeworld-api-0.3.1-unit-tests.log
    0 of 1 test suites (0 of 1 test cases) passed.

    ========== Aborting: Command failed. ==========
    DIRECTORY: codeworld-api
    COMMAND  : cabal test
    EXIT CODE: 1
    ===============================================

This was with the test suite, but not with your new flag or changes to CodeWorld.Driver, but these really shouldn't matter in this case.

cdsmith:
By the way, I understand why you needed to add the flag to `codeworld-api`, but we can't leave it there.  With that flag in place, running tests on the production server breaks the server, because it reinstalls a `codeworld-api` version that doesn't work.

Instead, I guess we need to duplicate the configuration from the library into the test suite, so that the same source files are rebuilt as part of the test suite itself.

nixorn:
previous problem with success tests resolved in https://github.com/google/codeworld/commit/0a30df5e510825e715a0200f42ca988ed7fd025c

"Instead, I guess we need to duplicate the configuration from the library into the test suite, so that the same source files are rebuilt as part of the test suite itself."

How to do this? Duplicate `codeworld-api` sources as `other-modules`?

cdsmith:
Right, you'd want to add both `test` and `src` in Cabal's `Hs-source-dirs` property, and merge the rest of the config over, as well.  So essentially, you're rebuilding the source code from `codeworld-api`as part of the unit test.  That's actually what we want, since we want to build it with different preprocessor flags.

The unfortunate part of that the config sections have to be duplicated in the cabal file.  This could probably be avoided with something like hpack, maybe?  But I think it's okay to just copy.

nixorn:
Ok. I'll copy, thanks.

cdsmith:
> previous problem with success tests resolved in 0a30df5

Okay, yes, I see.  You were calling `runTestTT` which returns counts of test failures, but then ignoring them and letting `main` succeed either way.  The simplest answer would have been to throw an exception if a test failed, but I like the solution of using `test-framework` instead.

nixorn:
@cdsmith I trying to mock browser, and it seems to going too deep.
for example some DOM-around functions here https://github.com/google/codeworld/pull/894/commits/ec2c7c0bf17bcfff9ccfe81abdfb319bece3035e#diff-3e471af93b2d4fa44b0cedd866cecbb8R872 . It is not all. Need to patch `canvas.getWidth`, things like `setAttribute` etc. It is going to become boilerplate explosion.

Either I following wrong way or there is better way to do this.

I think need to separate data and view. Functions which working with state (`stepHandler` and `eventHandler`) and drawing function, then pick state functions to tests.

cdsmith:
Yeah, that strikes me as the wrong way to do it.  Instead of trying to mock out the browser, I think what's wanted here is to move that stuff below an abstraction layer that hides the browser bits.

I don't know how to help you here.  This is a pretty sticky problem.  I think you've made good progress, but maybe it's time to shift gears again and look for more low-hanging fruit.

nixorn:
@cdsmith  I got it! I'm happy! :smile: https://github.com/google/codeworld/pull/894/files#diff-58d5580dae82e9a33e1a6745683014cbR22

nixorn:
At least part of problem.

nixorn:
btw `modifyMVarIfDifferent mv id` makes test pass.

cdsmith:
Sorry, I'm confused.

The last test you wrote should fail.  modifyMVarIfDifferent is intended to compare for pointer equality, because in general states cannot be compared for equality.  Your expression `\n -> n + 1 - 1` is semantically the identity function, but it doesn't preserve sharing, and your test proves that.  But this is all intended.  (Technically, you could argue it should be called unsafeModifyMVarIfDifferent instead, I guess)

The bug is that the event loop in `run` isn't preserving sharing.  We can't reproduce this without running the event loop.

nixorn:
Ok, place where we looking if next loop turn need time is https://github.com/google/codeworld/blob/master/codeworld-api/src/CodeWorld/Driver.hs#L1546
this https://github.com/google/codeworld/blob/master/codeworld-api/src/CodeWorld/Driver.hs#L1548 line is forever True, because we modify state variable here https://github.com/google/codeworld/blob/master/codeworld-api/src/CodeWorld/Driver.hs#L1540 (loop starts with `nextNeedsTime` == True)
In `runInspect` we wrap `eventHandler` and `stepHandler` into `DebugState`, so originally `(const id)` becomes something else(i guess), `fullStepHandler` forever modify value and `nextStateName` is forever `/= nextStateName`

nixorn:
I reproduced this by just commenting and rebuilding.

nixorn:
Other place is probably `isUniversallyConstant` as you noticed before. https://github.com/google/codeworld/blob/master/codeworld-api/src/CodeWorld/Driver.hs#L1124 `f undefined old` where 'f' ('fullStepHandler') will try to evaluate first argument will raise exception -> will become `False`

nixorn:
So I guess root of problem in wrapping of original state into debug state in `runInspect`, which turns each pointer-equality invariant into true equality. Maybe need replace wrapping by something like `DebugMonad`... I'm not sure.

cdsmith:
Okay, if I were you, I'd start writing some unit tests that test the pieces.  For example, you may want to check that `wrappedStep` preserves identity precisely when its argument function does.  You can write that test as follows:

    identical :: a -> a -> Bool
    identical !x !y = case reallyUnsafePtrEquality# x y of
        0# -> False
        _  -> True

    wrappedStepKeepsUnderlyingSharing :: Assertion
    wrappedStepKeepsUnderlyingSharing = do
        let wrapped = wrappedInitial 42
        assertBool "wrapping shared identity is shared" $
            identical wrapped (wrappedStep (const id) 1 wrapped)

        let unsharedId x = (x + 1) - 1
        assertBool "wrapping unshared identity is unshared" $ not $
            identical wrapped (wrappedStep (const unsharedId) 1 wrapped)

If you work your way down layers, your tests may fail at some point, and you'll know that's the point where the problem lies.

nixorn:
Thanks. I'm not figured out with `reallyUnsafePtrEquality#`. Just done with `makeStableName`. It is good. So now I can try to fix #681

cdsmith:
Sure, `makeStableName` and `reallyUnsafePtrEquality#` ultimately do the same thing: they observe reference equality.  So you can implement `identical` just as well like this:

    identical :: a -> a -> Bool
    identical !x !y = unsafePerformIO $ do
        nx <- makeStableName x
        ny <- makeStableName y
        return (nx == ny)

Does this test fail?  If so, then yes, let's fix it.  Otherwise, you probably want to add tests at different layers.  I definitely want to have a failing test case.  This has broken before, and I'd like to get a test in place that can detect when it happens so I don't have to reopen this bug every few months.

nixorn:
`wrappedStep` did not respect identity. Now it does :smile: But it freezes updating of `lastInteractionTime`. I'm not sure if this breaks something.

Pointer equality works well in pure functions even if we update the value,  but if we use `modifyMVar` + `readMVar` updating value to semantically same will change pointer. There are other places where happens update instead of identity return. I'm looking further.

nixorn:
`eventHandlerWrapper` from `runInspect` is first candidate.

nixorn:
I'm stuck with current situation. 

`samePointer` function checks stable name equality. I inserted this function into both [wrappedStep](https://github.com/google/codeworld/commit/95ea898267aa9e6d55e969446ddcb84dfabf8510#diff-3e471af93b2d4fa44b0cedd866cecbb8R1973) and [wrappedEvent](https://github.com/google/codeworld/commit/95ea898267aa9e6d55e969446ddcb84dfabf8510#diff-3e471af93b2d4fa44b0cedd866cecbb8R1998). With `wrappedStep` working as expected, but with `wrappedEvent` going on something weird.

```
Running 1 test suites...                                                                                                                                                 
Test suite unit-tests: RUNNING...                                                                                                                                        
Driver:                                                                                                                                                                  
True                                                                                    
  wrapping of shared identity is shared (steps): [OK]                                                                                                                    
True # debug print in wrappedEvent confirms names equality before and after applying handler to state
False # but same debug print in test says the opposite    
  wrapping of shared identity is shared (events): [Failed]
```

How is possible than pattern-matched wrapped state lose identity here https://github.com/google/codeworld/commit/95ea898267aa9e6d55e969446ddcb84dfabf8510#diff-3e471af93b2d4fa44b0cedd866cecbb8R1998 ?

cdsmith:
I'm merging this partial work, since you're planning to switch gears and look at UI stuff again.

Issue #339, opened 2016-09-08
Missing some important list blocks


cdsmith:
List blocks that I definitely would like to see because they are empirically useful:

* `repeating`
* `shuffled`
* `sorted`
* `reversed`
* `++`
* `first`

cdsmith:
I still need a `(++)` block, which seems more complex than the ones I've built so far.

PR #895, opened 2019-04-08
Adds ++ to list blocks

ZavierHenry:
Adds a ++ block referenced in #339


cdsmith:
Thanks!  Are you using the blocks UI?  It's definitely been somewhat neglected lately, so I'll pay more attention if it's being used again.

ZavierHenry:
Oh I just came across this project and thought it was interesting so I decided I'd try to fix some things that I could

PR #897, opened 2019-04-12
Closes #885

### LINKED ISSUE #885: Cancel button doesn't work when compiling
When compiling takes a long time, the Cancel button is grayed out and doesn't do anything.  It should actually cancel the compile.

ZavierHenry:
This bug occurred because `showLoading()` disables the cancel button when invoked


cdsmith:
Thanks for looking into this!

PR #900, opened 2019-04-16
Resolve 884

### LINKED ISSUE #884: Move editor documentation to the guide
The "Editor" button is too much screen emphasis for a minor reference.  It should be an appendix in the guide, instead.

nixorn:
Closes #884

### DIFF
--- a/web/help/editor.md
+++ b/web/help/editor.md
@@ -1,0 +1,32 @@
+Editor Shortcuts
+================
+
+    | `Shortcut`                   | `Action`                     |
+    |------------------------------|------------------------------|
+    | Ctrl + Enter                 | Run the program              |
+    | Ctrl + Space / Shift + Space | Autocomplete                 |
+    | Ctrl + Up                    | Zoom in                      |
+    | Ctrl + Down                  | Zoom out                     |
+    | Ctrl + A                     | Select all                   |
+    | Ctrl + Home                  | Go to start                  |
+    | Ctrl + End                   | Go to end                    |
+    | Alt + Left                   | Go to start of line          |
+    | Alt + Right                  | Go to end of line            |
+    | Ctrl + D                     | Delete line                  |
+    | Ctrl + Left                  | Go one word left             |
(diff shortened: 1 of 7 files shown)


web/help/editor.md
```diff
@@ -0,0 +1,32 @@
...
+    | Ctrl + Shift + F             | Replace                      |
+    | Ctrl + Shift + R             | Replace all                  |
+    | Ctrl + S                     | Save                         |
+    | Ctrl + Z                     | Undo                         |
+    | Ctrl + Shift + Z / Ctrl + Y  | Redo                         |
+    | Ctrl + U                     | Undo selection               |
+    | Ctrl + Shift +  U / Alt + U  | Redo selection               |
+    | Tab / Ctrl + ]               | Indent                       |
+    | Shift + Tab / Ctrl + [       | Unindent                     |
+    | Ctrl + I                     | Reformat (Haskell mode only) |
```
cdsmith: Since the keys are different between CodeWorld and Haskell modes, I think we actually want different documents for each mode.  So you'll need a `haskell.shelf` that will initially just contain the API docs and the editor help.  The editor help pages will be almost the same, but autocomplete is only in CodeWorld mode, and Reformat is only in Haskell mode.

web/help/editor.md
```diff
@@ -0,0 +1,32 @@
+Editor Shortcuts
+================
+
+    | `Shortcut`                   | `Action`                     |
+    |------------------------------|------------------------------|
+    | Ctrl + Enter                 | Run the program              |
+    | Ctrl + Space / Shift + Space | Autocomplete                 |
```
cdsmith: Just a nit, but this terse notation will confuse students.  I'd rather see:

    Ctrl + Space, or Shift + Space

Same with other keys below.

web/help/editor.md
nixorn: Done.

web/help/editor.md
nixorn: Done.

PR #901, opened 2019-04-16
WIP for docker build

bbarker:
Please do not merge yet.

The hope is that this will be, upon initial release, the container will be useful for both development and production, and will document the two different use cases. Since the development docker image may take much more storage than necessary for a runtime codeworld environment (though I'm not sure how much more, since codeworld still includes a ghc compiler), it is possible we might want to have two separate images in the future, but personally I'm not too worried about it.

Once working, we should think about under what repository to publish the images to: e.g., `https://hub.docker.com/u/google`, `https://hub.docker.com/u/bbarker`, or the (currently non existent) `https://hub.docker.com/u/codeworld`, or something else.

I'll be pasting some diagnostic output soon (once I have it again); the current situation though is that curl seems to hang when fetching mtl. I seemingly needed to alter the `wget` arguments to avoid a hang there as well, though I'm not sure if it is related.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ dist
 *.hi
 *.jsexe
 *.swp
+*~
 nohup.out
 web/closure-library
 third_party/closure-library
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM ubuntu:18.04
+MAINTAINER Brandon Barker <brandon.barker@cornell.edu>
+
+ARG GIT_COMMIT
+
+RUN apt update -y
+
+RUN apt install -y --no-install-recommends \
+    ca-certificates git sudo && \
+  mkdir /work && \
+  cd /work && \
+  git clone https://github.com/google/codeworld.git && \
+  cd codeworld && \
+  git checkout $GIT_COMMIT && \
+  ./install.sh
+
+RUN echo "#!/usr/bin/env bash\n\$@\n" > /opt/entrypoint && \
+  chmod a+x /opt/entrypoint
+
+ENTRYPOINT ["/opt/entrypoint"]
+
--- a/README.md
+++ b/README.md
@@ -88,6 +88,20 @@ If you make changes to CodeWorld, you can rebuild it without rebuilding the depe
 2. Run `./build.sh` to recompile just CodeWorld itself, using previously installed tools and libraries.
 3. Run `./run.sh` to start the server.
 
+Docker
+-------
+
+### To use the image from Docker Hub
+
+`TODO`
+
+### To build a new image
+
+`GIT_COMMIT=<commit> ./build-docker.sh`
+
+or just `./build-docker.sh` to use the commit that is currently
+checked out in your local repository.
+
 Caveats
 -------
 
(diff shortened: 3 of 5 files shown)


bbarker:
Here is the current failure point:

```
===== IN /work/codeworld/build, RUNNING: tar xf /work/codeworld/build/downloads/cabal-install-2.0.0.0.tar.gz
===== IN /work/codeworld/build/cabal-install-2.0.0.0, RUNNING: ./bootstrap.sh
Using gcc for C compiler. If this is not what you want, set CC.
Using /usr/lib/gcc/x86_64-linux-gnu/7/collect2 instead.
Checking installed packages for ghc-8.0.2...
deepseq is already installed and the version is ok.
binary is already installed and the version is ok.
time is already installed and the version is ok.
transformers is already installed and the version is ok.
mtl-2.2.1 will be downloaded and installed.
text-1.2.2.2 will be downloaded and installed.
parsec-3.1.11 will be downloaded and installed.
network-2.6.3.2 will be downloaded and installed.
old-locale-1.0.0.7 will be downloaded and installed.
old-time-1.1.0.3 will be downloaded and installed.
HTTP-4000.3.7 will be downloaded and installed.
zlib-0.6.1.2 will be downloaded and installed.
random-1.1 will be downloaded and installed.
stm-2.4.4.1 will be downloaded and installed.
async-2.1.1.1 will be downloaded and installed.
base16-bytestring-0.1.1.6 will be downloaded and installed.
base64-bytestring-1.0.0.1 will be downloaded and installed.
cryptohash-sha256-0.11.100.1 will be downloaded and installed.
mintty-0.1 will be downloaded and installed.
echo-0.1.3 will be downloaded and installed.
edit-distance-0.2.2.1 will be downloaded and installed.
ed25519-0.0.5.0 will be downloaded and installed.
tar-0.5.0.3 will be downloaded and installed.
hashable-1.2.6.1 will be downloaded and installed.
hackage-security-0.5.2.2 will be downloaded and installed.
Cabal-2.0.0.2 will be downloaded and installed.

Downloading Cabal-2.0.0.2...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:05:00 --:--:--     0
curl: (28) Operation timed out after 300682 milliseconds with 0 out of 0 bytes received

Error during cabal-install bootstrap:
Failed to download Cabal.
```

It appears to mimic https://github.com/moby/moby/issues/37878 but there's no VPN involved on my end here. Would be interesting to know if others have the same issue though.

nixorn:
I'm not sure docker required in production, but I guess shared docker image could prevent situations like this https://github.com/google/codeworld/issues/975

I think main point of dockerization is possibility to play with local codeworld installation without doing expensive `install.sh` (on my laptop with my connection it takes several hours).

@bbarker Hi. Are you still on this? If no, can I take this?

cdsmith:
Unfortunately, it's likely that this pull request is a bit obsolete.  I've revamped a bunch of the installation procedure as part of the update to GHCJS 8.6.

cdsmith:
I'll close this, since we now have a docker file that works.

PR #902, opened 2019-04-17
Toolbar with naive clipboard.

### LINKED ISSUE #883: Toolbar in the CodeWorld editor
The CodeWorld editor could use a toolbar with buttons for cut, copy, paste, undo, save, and other basic editor features.  Many students know the standard icons for these things, but don't know key combinations.

nixorn:
Closes #883

### DIFF
--- a/third_party/MaterialDesign/content-copy.svg
+++ b/third_party/MaterialDesign/content-copy.svg
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   version="1.1"
+   width="24"
+   height="24"
+   viewBox="0 0 24 24"
+   id="svg4"
+   sodipodi:docname="content-copy.svg"
+   inkscape:version="0.92.3 (2405546, 2018-03-11)">
+  <metadata
+     id="metadata10">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+      </cc:Work>
+    </rdf:RDF>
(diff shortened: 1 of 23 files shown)


web/js/codeworld.js
```diff
@@ -218,7 +220,68 @@ function initCodeworld() {
...
+                    window.clipboard = cm.getSelection();
+                    document.execCommand('copy');
+                }
+            },
+            {
+                class: 'cw-button blue cw-button-paste',
+                label: '',
+                title: 'Paste',
+                callback: function (cm) {
+                    cm.replaceSelection(window.clipboard);
```
cdsmith: Does this mean that in addition to the system clipboard, you are keeping your own clipboard inside the web page?  That's going to cause lots of problems, such as making it hard to copy from another tab, an email, etc.  Any chance there's a way to request permissions we need and do this the right way?  Otherwise, maybe it's better to leave these off...

nixorn:
Screencast 
Peek 2019-04-18 14-14

web/js/cm_addons/codemirror-buttons/buttons.js
```diff
@@ -0,0 +1 @@
+../../../../third_party/codemirror-buttons/buttons.js
```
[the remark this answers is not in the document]
nixorn: Ok, I will move.

web/js/codeworld.js
nixorn: useful resources which contains drafts how to do this https://www.lucidchart.com/techblog/2014/12/02/definitive-guide-copying-pasting-javascript/

https://jsfiddle.net/vtjnr6ea/ - looks like working example

I'm in progress with this.

nixorn: I can copy to clipboard, but can not copy from. According to this https://caniuse.com/#feat=clipboard we can not use copying from clipboard  via paste command in firefox. I can listen for `paste` event but not sure I can throw custom paste event on button click.

third_party/MaterialDesign/content-copy.svg
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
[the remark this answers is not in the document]
nixorn: Already here https://github.com/google/codeworld/blob/master/third_party/MaterialDesign/license.txt .

web/js/codeworld.js
nixorn: Correction firefox should support paste with a focused editable field. In progress.

nixorn: I have no success with copying from system clipboard. Should I search for solution, or just delete "Past" button?

cdsmith:
Unfortunately, I had to revert this because it caused the editor scrolling to become unusable.  Can you take another look?  The revert commits are 8cec55e2d1516d8f648ae686631a4ec723af0bcc (for your code) and 1bd068f22a70788310bc3cad067972e482934fc3 (for styling I did after the merge)

nixorn:
Ok, fixing this.

PR #904, opened 2019-04-20
Tooltip pane

### LINKED ISSUE #883: Toolbar in the CodeWorld editor

nixorn:
closes #883

Previous attempt broke scrolling due to change editor position from `absolute` to `relative`. But just return to `absolute` and moving editor down was not enough, addons/display/panel.js (requirement for buttons.js) contain own resizing logic for editor field, which looks bad.

I threw away panel.js, and took from it only several lines we really need.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -63,7 +63,6 @@ function build_codemirror {
       addon/dialog/dialog.js \
       addon/display/placeholder.js \
       addon/display/rulers.js \
-      addon/display/panel.js \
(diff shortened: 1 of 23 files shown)


cdsmith:
Thanks.

PR #908, opened 2019-04-23
Add a cooldown to the Asteroids example

josephcsible:
Currently, when holding the up arrow while out of energy, not only do you
not get any thrust, but you also don't recover any energy. Fix this by
turning off thrust when energy runs out, until it either fills all the way
back up, or the player releases the key.

### DIFF
--- a/web/help/examples.md
+++ b/web/help/examples.md
@@ -375,6 +375,7 @@ data World = World {
     left         :: Number,
     right        :: Number,
     thrust       :: Number,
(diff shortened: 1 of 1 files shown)


josephcsible:
This is a different solution to the issue I originally tried to fix in #809.

cdsmith:
I'm merging this because you seem to care about this example more than I.  To be honest, though, I think all the examples are too complex to be useful.

PR #909, opened 2019-04-23
Closes #886

### LINKED ISSUE #886: codeworld-auth is incompatible with JWT 0.10

ZavierHenry:
Quick note: Because the signature of `encodeSigned` changes from `Signer -> JWTClaimsSet -> JSON` in previous JWT versions to `Signer -> JOSEHeader -> JWTClaimsSet -> Text` in JWT 0.10, updating to JWT 0.10 means not being able to use previous JWT versions in the cabal file


cdsmith:
Thanks

PR #911, opened 2019-04-24
Dark mode for codeworld and haskell modes

### LINKED ISSUE #877: Dark mode
Would be nice to have a dark theme.

nixorn:
Closes  #877
It is just inverting of all colors + enabling theme for CodeMirror. In some cases it looks bad.

I working on enabling dark mode for doc iframe and maybe runner iframe. After that need to tweak colors for dark mode.

Funblocks untouched.

### DIFF
--- a/web/css/codeworld.css
+++ b/web/css/codeworld.css
@@ -398,6 +398,7 @@ body {
     position: fixed;
     top: 3vh;
     left: 10vw;
+    background-color: var(--helpdoc-background-color);
 }
 
 .swal2-popup.helpdoc * {
--- a/web/css/codeworld.css
+++ b/web/css/codeworld.css
@@ -31,7 +31,7 @@ body {
     -webkit-align-items: stretch;
     -ms-align-items: stretch;
     height: 100vh;
-    background-color: #e5eff9;
+    background-color: var(--main-bg-color);
     display: flex;
     display: -webkit-flex;
     display: -ms-flex;
@@ -128,7 +128,7 @@ body {
 }
 
 #keyboard-shortcuts table tbody tr td {
-    border: 1px solid black;
+    border: 1px solid var(--kb-shortcuts-table-border);
     padding: 4px;
 }
 
@@ -187,7 +187,7 @@ body {
 }
 
 #message.error {
-    background-color: #fcc;
+    background-color: var(--message-bg-color);
 }
 
 .message-box {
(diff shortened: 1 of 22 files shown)


nixorn:
Peek 2019-04-24 16-23

cdsmith:
Looks like a great start.  I agree that the colors need some fine-tuning.  You might also consider removing the text from the button, to make it take up less space.  Screen space is at a premium on small laptops.  If you do so, please add a title attribute for accessibility.

nixorn:
Should we do dark theme for help pane? If no, this is ready for review. I updated acidic colors for buttons, and fixed styles for links.

nixorn:
Also if this going to be merged probably we should wrap any css color we use into variable, to be possible customize it for more than one color theme.

nixorn:
Hm, did not checked with haskell mode. Haddock docks looks terrible. Fixing.

nixorn:
Much better now. 
Peek 2019-04-29 16-44

web/css/doc.css
```diff
@@ -30,16 +30,16 @@ body {
...
     padding: 3px;
-    border: dotted #888 1px;
+    border: dotted var(--doc-popuout-a-border) 1px;
     border-radius: 5px;
 }
 
-.md pre.tilde.clickable {
-    background-color: #f0f8ff;
+.md pre.tilde {
+    background-color: var(--doc-tilde-bg) !important;
```
cdsmith: Use of `!important` is sometimes bad for accessibility, since it overrides customizations in user style sheets.  Can you get this to take priority by being more specific, instead?

Same for other uses below.

web/css/codeworld.css
```diff
@@ -502,11 +507,11 @@ body {
...
     pointer-events: none;
     background: linear-gradient(to bottom,
-        rgba(0, 0, 0, 0.7) 0px,
-        rgba(0, 0, 0, 0.6) 2em,
-        rgba(0, 0, 0, 0) 3em,
-        rgba(0, 0, 0, 0) 100%);
-    color: white;
+        var(--obsolete-code-alert-grad-point1) 0px,
+        var(--obsolete-code-alert-grad-point2) 2em,
+        var(--obsolete-code-alert-grad-point3) 3em,
```
cdsmith: The third and fourth color points will always be the same, and will always be 100% transparent.  You don't need variables for those.

web/css/doc.css
```diff
@@ -128,3 +134,43 @@ body {
...
+.ui-widget-content a {
+    color: var(--doc-accordion-color);
+}
+
+.md svg.diagram {
+    fill: var(--doc-svg-fill) !important;
+    stroke: var(--doc-svg-stroke) !important;
+}
+
+/* Haskell-mode. */
```
cdsmith: I'm not sure what this comment means, but nothing here is specific to haskell mode.  Wouldn't it also apply to the reference sections of codeworld-mode?

web/env.html
```diff
@@ -110,6 +112,7 @@
                 <a id="helpButton" class="cw-button blue" target="_blank" href="http://help.code.world"><i class="mdi mdi-18px mdi-comment-question-outline"></i>&nbsp;
                     Ask a Question</a>
                 <a id="bugButton" class="cw-button blue" target="_blank" href="https://github.com/google/codeworld/issues/new?template=bug_report.md"><i class="mdi mdi-18px mdi-bug"></i>&nbsp; Report a Bug</a>
+                <a id="toggleThemeButto" class="cw-button blue theme-marker-day" title="Toggle theme" target="_blank" onclick="toggleTheme()"><i class="mdi mdi-18px mdi-weather-sunny"></i></a>
```
cdsmith: Misspelling in the `id` atttribute.  It doesn't look like you even need the id anyway.

nixorn: True. Deleting this.

web/css/doc.css
nixorn: In places where `!important` used I override inline styles applied by `markdeep`. This happens after loading of doc.css, and have higher priority.
What means more specific?

web/css/doc.css
nixorn: Deleted.

web/css/doc.css
cdsmith: Okay, that makes sense.  What I meant was that it's possible to make one CSS selector override another by making it more specific.  For example, a `div.foo` rule will always take precedence over a  `foo` rule, if they give different values for the same property.

Unfortunately, though, there is no way to take precedence over an inline style on an HTML element.  It's always considered the most specific rule, and take priority over any selector.  So use of `!important` here is the right choice.  Thanks for explaining.

3noch:
So nice! Thank you!

PR #912, opened 2019-04-26
Fix LocalAuth.

nixorn:


web/env.html
```diff
@@ -185,6 +185,7 @@
     <script type="text/javascript" src="js/details-element-polyfill.js"></script>
     <script type="text/javascript" src="js/codeworld-mode.js"></script>
     <script type="text/javascript" src="js/codeworld_shared.js"></script>
+    <script type="text/javascript" src="js/codeworld_local_auth.js"></script>
```
cdsmith: This should not be needed.  This script is loaded at https://github.com/google/codeworld/blob/2c43b42264cfbad4554151a99c8cbcb14d158b16/web/js/codeworld_shared.js#L577

If it's not getting loaded for you, let's figure out why not.  But hard-coding local auth to always be loaded won't work, because we don't use local auth in the production server.

cdsmith: Looking over the code, in order for local auth to be configured, you need:

1. web/clientId.txt should NOT exist
2. codeworld-auth.db and codeworld-auth.txt SHOULD exist.

There are instructions for creating and managing these files at https://github.com/google/codeworld#local-auth

nixorn: Ok, thanks, got it. Investigating.

PR #915, opened 2019-05-04
(&) keeps shadow parameters when mutator is opened

### LINKED ISSUE #309: (&) loses its shadow parameters when the mutator is opened
The (&) operator starts with shadow blocks for its picture parameters.  But as soon as the config dialog is opened, they disappear.  They should stay.

ZavierHenry:
Closes #309


cdsmith:
Thanks!

PR #916, opened 2019-05-06
correct documentation of 'dilated'

jvoigtlaender:
The documentation for `dilated` currently reads:
```
A picture scaled by these factors.
```
The plural in there is strange, as `dilated` only takes one factor as argument. So the documentation should read:
```
A picture scaled by this factor.
```


cdsmith:
Thanks for the report.  Rather than asking you to sign a CLA, I've rewritten the documentation myself to fix the issue.

cdsmith:
Oops; fixed in commit 921ce29c0f8511c8d9fd93895bb3642dbdbe7e46, but I mistyped the issue number.

jvoigtlaender:
Thanks. Very welcome. I was wondering whether to do the CLA for such a small issue.

PR #942, opened 2019-05-20
Adds missing RGB and HSL in blocks parser

### LINKED ISSUE #933: RGB and HSL do not work in blocks
**Shared Program**: https://code.world/blocks#PvRj8UPGBBE_9FNsQRouwzQ

**Description**
In the shared program above, the RGBA block does produce a second parameter to the colored function, but substituting RGB does not create a second parameter.

ZavierHenry:
Closes #933


cdsmith:
Thanks for the fix

Issue #941, opened 2019-05-20
clipped: clipping an existing image


cdsmith:
As I was considering what it would mean to have a conceptually infinite `coordinatePlane`, it occurred to me that `clipped :: (Picture, Number, Number) -> Picture` would be useful, where `clipped(pic, w, h)` is `pic` inside a central w-by-h rectangle, and transparent elsewhere. 
 Implementation is easy, using https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/clip.

This fills some of the more trivial use cases of #28 but without the semantic complexity or implementation difficulties.

alphalambda:
The `clipped` function was in my wish list, but I would like to be able to clip along an arbitrary path, so that I can *punch* holes in pictures and create textures that do not *bleed*. Maybe we could have a  simple rectangular `clipped` and an advanced `clippedAlongPath` version.

cdsmith:
Hmm.  If this gets too complex, then I think I'd prefer a more general-purpose API like #28 instead of a bunch of special cases, at least in the default API.  Constructive geometry is a much better approach to things like cutting holes out of pictures, etc.

alphalambda:
I have not thought about it enough to grasp whether implementing clipping along a path would be as complex as implementing constructive geometry or not. The rest of my comment here applies only in case implementing `clipped` is worth it.

If that is the case, then `clipped` would probably be implemented in `codeworld-api` in terms of an internal `clippedAlongPath` function anyway. My only request is that `clippedAlongPath` is exported from `codeworld-api` even if it is not included in `codeworld-base.Prelude`.

cdsmith:
The implementation for clipping along a path is definitely easier.  I'm concerned that as use cases are added, the interface will explode in complexity.  The key question is this: what is a clipping region, really? Is it good enough to clip to a straight-line path?  Probably not.  The natural next thing to ask for is clipping along bezier curves, or circles.  CodeWorld doesn't have a "path" or "region" type to represent these options.

Up until now, CodeWorld has been able to simplify a lot by insisting that there's only one Picture type representing figures on the screen.  This contrasts with other libraries, which often have different types for paths, shapes, figures, collages, etc.  In that spirit, it would be great if we could somehow reuse Picture as the type to represent a clipping region.  But then clipping is really just `&` with different alpha compositing rules, and that gets right back to the constructive geometry angle.

cdsmith:
[...]

Issue #938, opened 2019-05-17
Support for arbitrary aspect ratios in output window


alphalambda:
I wonder how difficult it would be to have support for arbitrary aspect ratios in the output, because a square output limits student possibilities to create interesting scenes, since most devices are actually not square. Restricting the window to be square forces them to waste a lot of space for their scene in the horizontal direction, especially when compared with similar environments that target the same population.

One of the things I like the most about CodeWorld is the fact that the output uses abstract display units instead of pixels or other device-dependent units. Also, I think 10x10 is a good default choice for a general drawing. However, I don't see why choosing to build a 15x10 scene would be a problem, especially now that there are built-in controls to zoom and pan the output.

The only possible problem could be the `coordinatePlane` facility, which I think should be a function anyway, so that the span and resolution are configurable. This could be also important for using CodeWorld as a programmable graphing calculator, for which it has a lot of potential.

cdsmith:
The reason for a fixed-size canvas has always been simplicity.  I agree with most of what you're saying, but would just need to see what the UI looks like.  Adding arguments to a function is I think a big step in complexity: `drawingOf(flower)` can be read and pronounced naturally, while `drawingOf(flower, 10, 10)` feels a lot more technical.

Maybe there's a clever way to do this without making the code complex.  Maybe even something in the UI, where you can select a device to build for outside of the code, and that includes the aspect ratio.

alphalambda:
I don't think that `drawingOf` needs to be modified. On the CodeWorld language side, the only modification I think is needed is adding a `wideCoordinatePlane` that takes in a `width` as a parameter, so that `wideCoordinatePlane(width)` fills a rectangular area of `width` by 20 units.

Then, in the GUI side, maybe a small button to *unlock* the aspect, so that when you move the left side of the output window while it is unlocked, it changes only the width. 

Finally, for the fullscreen mode, maybe a parameter in the URL, like `... mode=codeworld&wide=30`.

I think that is all that is needed, as a drawing is potentially infinite, so it is just a convention that you create the interesting part of your drawing so that it fits the expected output.

cdsmith:
Okay, this sounds fine.  Although I might prefer one coordinatePlane Picture value that is conceptually infinite (of course, the drawing code would look at the transformations and decide how much needs to be drawn).  This might be unfortunate for current projects that rely on the bounded coordinate plane, though.

alphalambda:
[...]

PR #943, opened 2019-05-20
Graphed now takes two separate scaling factors

alphalambda:
When I showed the new graphed function to the teacher, she immediately requested being able to specify the X and Y scales separately. So, I changed graphed to accommodate that. This is an incompatible change, but nobody should be using it yet, so I guess it is OK.

While I was at it, I also implemented wideGraphed as a proof of concept. I don't want to step on anybody's toes, but I would be happy to go ahead and implement wideCoordinatePlane if you want.

Here is a link to the code that the teacher did immediately after I provided her with the interface:

https://blackeyepeas.phys.lsu.edu/cwn/#Pf93EdrftpZNnpiWXJi4BBQ

### DIFF
--- a/codeworld-base/src/Extras/Cw.hs
+++ b/codeworld-base/src/Extras/Cw.hs
@@ -549,7 +549,7 @@ underlays(f,n) = underlays'(f,max(0,truncation(n)))
 --
 -- Example 1 shows a circle that grows forever, while the graph keeps
 -- adjusting the scale so that it fits within the output.
--- Example 2 shows a graph with 100 circles. It used the function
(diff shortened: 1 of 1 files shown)


cdsmith:
Thanks for the quick fix!  I had not pushed your earlier change to production yet, so I can guarantee no one used it there.

If you want to add `wideCoordinatePlane` in Extras, go for it. I'm considering you the owner of those packages for now, so you're free to add whatever you like there.  As far as the coordinate plane issues discussed in #938 and #941, I'd like to think a bit before committing to something in the core API there.

PR #949, opened 2019-05-26
Extras.Colors

alphalambda:
Moved the color convenience functions from Extras.Cw to Extras.Colors. Probably very few people, if anyone, were using them, so I better do it now before my new teacher training begins.
I also added a `colorNamed` function that takes X11 color names and also hexadecimal strings. I wanted
to try this with the new cohort, because it seems that I was wrong, and students really like to use color
names. But I would not like to pollute the variable space with lots of pre-defined color variables. I want to
test how this trade-off of using color names works. Yes, it is less safe, but it is just colors, anyway...
And this way CodeWorld students will be able to use the same colors they use in HTML or in other programs.

### DIFF
--- a/codeworld-base/src/Extras/Colors.hs
+++ b/codeworld-base/src/Extras/Colors.hs
@@ -32,7 +32,7 @@ painted(pic,name) = colored(pic,colorNamed(name))
 -- | This function allows you to specify color components in the range 0 to 255
 -- instead of 0 to 1.
 rgb :: (Number,Number,Number) -> Color
-rgb(r,g,b) = RGB(r/256,g/256,b/256) -- Not exact, but colors aren't anyway...
+rgb(r,g,b) = RGB(r/255,g/255,b/255)
 
 -- | A shade of grey as given by the argument, where @grey(0)@ is @black@
 -- and @grey(1)@ is white.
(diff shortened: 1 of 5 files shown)


codeworld-base/src/Extras/Colors.hs
```diff
@@ -0,0 +1,271 @@
...
+
+-- | The given picture painted with the color corresponding to the given
+-- color name, which must be a valid argument of the 'colorNamed' function.
+painted :: (Picture,Text) -> Picture
+painted(pic,name) = colored(pic,colorNamed(name))
+
+-- | This function allows you to specify color components in the range 0 to 255
+-- instead of 0 to 1.
+rgb :: (Number,Number,Number) -> Color
+rgb(r,g,b) = RGB(r/256,g/256,b/256) -- Not exact, but colors aren't anyway...
```
cdsmith: Shouldn't this be dividing by 255 instead of 256?

alphalambda: Yes. It should be 255. That's what the *not exact* comment in the code was about.  I will correct it.

I just thought: 255 is the worst possible divisor, while 256 is the best, and there is very little visible difference...

PR #951, opened 2019-05-26
Fixed bug in the URL generated by Share without code when the server is not at /

alphalambda:
This PR is not for Extras. It fixes the generated URL to add the correct pathname for *Share without code*.


cdsmith:
I don't think this will work correctly for https://code.world/haskell  You're likely to get a URL like https://code.world/haskellrun.html?...

However, it looks like just removing the slash should work.  I'll make that change instead.

PR #956, opened 2019-05-28
Tree interface for projects tool

### LINKED ISSUE #827: Revamp project handling UI
The current project handling UI grew organizally from whatever happened to already be there, and people find it confusing.  It should be replaced by something that acts more intuitively.  Some improvements would be:

1. A tree view whose parts can be expanded and collapsed at will, instead of having only one currently expanded path.
2. Reordering and moving projects by drag and drop (with perhaps a sort button that puts the projects back in order, by time or alphabetically?).
3. A search or filter bar to look for projects based on substrings of the name.

All of this fits into the current model (with some added metadata for ordering).  It's just a client-side revamp.

nixorn:
closes #827 

Peek 2019-05-28 17-47

It is draft, there are lot of things to do yet.
1. Make same for funblocks (currently in this bramch funblocks is broken)
2. Implement sorting. I going to do it by inserting `order` Int field to `dir.info` and source metadata
3. cleanup(both code and commit history), testing etc.


cdsmith:
This pull request **introduces 3 alerts** when merging 2aa8e909b2a46b05fcedf4a12a84a927af101111 into b99f09d5437a4fc86bf58fb5f5ea0369d298aab2 - [view on LGTM.com](https://lgtm.com/projects/g/google/codeworld/rev/pr-832d6c3a28065d33b215a324029968818cf0a2d0)

**new alerts:**

* 3 for Superfluous trailing arguments

---

*Comment posted by [LGTM.com](https://lgtm.com)*

cdsmith:
Okay, let me propose another direction: what about completely replacing the existing project browser with something based on https://www.jstree.com/ ???

This accomplishes some of what is needed, but it is still based on buttons instead of familiar-looking controls, which was a big part of the bug.  It doesn't implement drag and drop.  It doesn't implement sorting.  If CodeWorld were to use jstree instead of reinventing its UI with buttons, all of these things would be available automatically.

What do you think?

nixorn:
I used https://mbraak.github.io/jqTree/ . I just wrote codeworld-like theme for it, it was relatively easy (around 150 LoC). jqTree have neat and minimalistic api. Jqtree have no in-box search and rename plugins, but really easy to make them via customizable node html.   

I tried jstree and I don't like it. It looks like legacy project. Default jstree theme with 1000+ lines of css scary me https://github.com/vakata/jstree/blob/master/src/themes/default/style.css .

nixorn:
"It doesn't implement drag and drop." - It is.
Peek 2019-05-29 11-42

'It doesn't implement sorting." - sorting not implemented on backend. We should save sorted state between user sessions to show files in order in which user leaved them last time. For that we need someting like `order` field in directory and project metadata. After that sorting will easy to implement. I'm working on it.

nixorn:
Drag and drop looks bad now, because after dropping I re - ask server about directory structure, and rerednering of all tree collapse it. I'll will fix that.

cdsmith:
Okay, great!  Glad to hear it

cdsmith:
This pull request **introduces 5 alerts** when merging c8f2f32fa1f55121e757c72c749413e4af47c853 into c963005988903db28c8e87641ca1ed97a5f032f8 - [view on LGTM.com](https://lgtm.com/projects/g/google/codeworld/rev/pr-30f7b82aabe096312e8a61ece28c489f819585ea)

**new alerts:**

* 5 for Superfluous trailing arguments

---

*Comment posted by [LGTM.com](https://lgtm.com)*

cdsmith:
This pull request **introduces 5 alerts** when merging ec67fc24bce28b73dc6fa922d0f6499e6ebf4016 into c963005988903db28c8e87641ca1ed97a5f032f8 - [view on LGTM.com](https://lgtm.com/projects/g/google/codeworld/rev/pr-0382edae7dacdfedd4450c33e3f78b32a6f51f9f)

**new alerts:**

* 5 for Superfluous trailing arguments

---

*Comment posted by [LGTM.com](https://lgtm.com)*

nixorn:
Peek 2019-06-02 14-13
This one ready for review. There is big refactoring with moving all project tree logic from `window.*` to new tree view. User defined order are supported.

cdsmith:
This pull request **introduces 1 alert** when merging 0fcf7b79fec10e15276d65f9ad688bf2c9fd7215 into c963005988903db28c8e87641ca1ed97a5f032f8 - [view on LGTM.com](https://lgtm.com/projects/g/google/codeworld/rev/pr-83c7c2feefdfaecde439daf0ddb98bae4f63efde)

**new alerts:**

* 1 for Superfluous trailing arguments

---

*Comment posted by [LGTM.com](https://lgtm.com)*

nixorn:
Hm, sorry, looks `format.sh` produced bigger diff. It could make review harder.

nixorn:
Also I did not test sharing, because I can not test sharing locally.

cdsmith:
Thanks.  I'll merge this when I have a chance to do the testing.  Don't worry about the formatting.  If it needed fixing, then so be it.

codeworld-server/src/Main.hs
```diff
@@ -152,7 +152,8 @@ site ctx =
             [ ("loadProject", loadProjectHandler ctx)
             , ("saveProject", saveProjectHandler ctx)
             , ("deleteProject", deleteProjectHandler ctx)
-            , ("listFolder", listFolderHandler ctx)
+            , ("directoryTree", directoryTreeHandler ctx)
+            , ("dumpTree", dumpTreeHandler ctx)
```
cdsmith: If I understand correctly, the point of `dumpTree` is to store an ordering of the files and folders in a directory, so that they can be reordered?  This seems like a very heavyweight way to accomplish that goal.  Instead of serializing the entire directory structure to disk in a single file, why not store just the ordering of items in the current folder, inside `dir.info` for that directory?

The main advantage here is that it's easy to combine this with the existing directory tree.  The ordering is used only to sort lists, and everything else can be deduced from the directory structure itself.  If the directory structure doesn't match, it's okay.  You know to trust the directory structure, and just stick any items with no existing sort order at the end, or something like that.

cdsmith: Whether you change the structure of the file or not, I'd really like to see this endpoint named something like `writeProjectOrder`.  To me, `dumpTree` sounds like a debugging endpoint, and it was really not obvious what this is for.

codeworld-server/src/Main.hs
```diff
@@ -253,19 +254,19 @@ deleteProjectHandler = private $ \userId ctx -> do
             projectFile projectId
     liftIO $ removeFileIfExists file
 
-listFolderHandler :: CodeWorldHandler
```
cdsmith: Can you leave the `listFolder` endpoint in place at first?  I can delete it later.  This way, I can update the server without breaking clients that have old JavaScript.

codeworld-server/codeworld-server.cabal
```diff
@@ -50,6 +50,7 @@ Executable codeworld-server
     SafeSemaphore,
     snap-core,
     snap-server,
+    sort,
```
cdsmith: You don't need this.  Just use `sort` from `Data.List` in `base`.

codeworld-server/src/Util.hs
```diff
@@ -296,3 +299,53 @@ removeDirectoryIfExists dirName =
     handleExists e
         | isDoesNotExistError e = return ()
         | otherwise = throwIO e
+
+loadDumpedTree :: BuildMode -> UserId -> IO DirTree
+loadDumpedTree bm uid = do
```
cdsmith: The convention so far is to use `mode` for variables containing a `BuildMode`.

codeworld-server/src/Util.hs
```diff
@@ -296,3 +299,53 @@ removeDirectoryIfExists dirName =
...
+    let file = userProjectDir bm uid </> "tree.info"
+    exists <- doesFileExist file
+    realTree <- getDirectoryTree bm uid
+    case exists of
+        False -> return realTree
+        True -> do
+            content  <- LB.readFile file
+            return $ case decode content of
+                Nothing -> realTree
+                Just dumpedTree -> case compareTrees realTree (Dir "root" dumpedTree) of
```
cdsmith: If you don't replace this entirely (though I think you should), I think you need to do something different here.  This seems to have the consequence that one unexpected file causes the entire directory tree to forget its sort order.

codeworld-server/src/Util.hs
```diff
@@ -296,3 +299,53 @@ removeDirectoryIfExists dirName =
...
+                    True  -> Dir "root" dumpedTree
+                    False -> realTree
+
+getDirectoryTree :: BuildMode -> UserId -> IO DirTree
+getDirectoryTree bm uid = do
+    ftr <- getDirectory $ userProjectDir bm uid
```
cdsmith: I'm not clear on what `ftr` means.  More descriptive name, please?

codeworld-server/src/Util.hs
```diff
@@ -296,3 +299,53 @@ removeDirectoryIfExists dirName =
...
+    children' <- mapM (constructTree rootPath) $ filter notInfo children
+    return $ Dir "root" children'
+    where
+        constructTree :: FilePath -> Tree FilePath -> IO DirTree
+        constructTree prefix (Node path [(Node "dir.info" [])]) = do -- empty directory
+            name <- B.readFile $ prefix </> path </> "dir.info"
+            return $ Dir (T.decodeUtf8 name) []
+        constructTree prefix (Node path []) = do
+            let currentNode = prefix </> path
+            source <- LB.readFile currentNode
```
cdsmith: So you're going to read every single user project into memory just to read a directory?  That seems excessive.  Why not just remember the directory structures, and the project names, so you can load them when they are needed later?  There is already a `loadProject` endpoint that does that.

codeworld-server/src/Util.hs
```diff
@@ -296,3 +299,53 @@ removeDirectoryIfExists dirName =
...
+    ftr <- getDirectory $ userProjectDir bm uid
+    let (Node rootPath children) = toTree ftr
+    children' <- mapM (constructTree rootPath) $ filter notInfo children
+    return $ Dir "root" children'
+    where
+        constructTree :: FilePath -> Tree FilePath -> IO DirTree
+        constructTree prefix (Node path [(Node "dir.info" [])]) = do -- empty directory
+            name <- B.readFile $ prefix </> path </> "dir.info"
+            return $ Dir (T.decodeUtf8 name) []
+        constructTree prefix (Node path []) = do
```
cdsmith: So it's only because of `dir.info` that this doesn't match an empty folder?  That seems like round-about logic.  A more direct way to decide would be to look at the filename: directories start with capital D, and projects start with S.

web/blocks.html
```diff
@@ -26,6 +26,7 @@
     <link rel="stylesheet" href="css/codeworld.css">
     <link rel="stylesheet" href="css/show-hint.css">
     <link rel="stylesheet" href="css/codeworld-cm.css">
+    <link rel="stylesheet" href="css/directoryTree.css" />
```
cdsmith: [nit] Change this everywhere else, if you do it here.

web/js/codeworld.js
```diff
@@ -35,22 +35,19 @@ async function init() {
...
+
     window.openProjectName = null;
-    window.nestedDirs = [''];
 
     window.savedGeneration = null;
     window.runningGeneration = null;
 
     if (window.location.pathname === '/haskell') {
         window.buildMode = 'haskell';
+        window.projectEnv = 'haskell';
```
cdsmith: Can you explain why you need this to be different from window.buildMode?

codeworld-server/src/Main.hs
nixorn: Ok, I'll do this.

nixorn: If we going to store order inside of `dir.info`, this endpoint will have different logic or even not required.
If it will be required after I'll rewrite this solution, I'll rename it into `writeProjectOrder`.

codeworld-server/src/Main.hs
nixorn: ok

codeworld-server/codeworld-server.cabal
nixorn: Ok, thanks.

codeworld-server/src/Util.hs
nixorn: Ok.

codeworld-server/src/Util.hs
nixorn: I'll rename it to `fileTree`

web/js/codeworld.js
nixorn: Funblocks build using `codeworld` mode, but sources are in `data/blocklyXML/projects` folder. So I used different variable to separate build mode and place from which need to load projects between `codeworld.js` and `funblocks.js`.

codeworld-server/src/Util.hs
nixorn: Okay, lazy loading is point I missed here.

In this solution user make one directory tree request per session and copy whole structure to js. After that changes to tree (moving, deleting, creating of nodes etc) becomes after 200 response from server, browser does not re ask all tree from server until next session.

I need get name of projects, but I can not do it without loading and decoding. Storing all tree in browser makes search pure js task without server requests.

How much entries in biggest projects directory in production?

codeworld-server/src/Util.hs
nixorn: `Node path [(Node "dir.info" [])]` match empty folder without projects. http://hackage.haskell.org/package/filesystem-trees-0.1.0.6/docs/System-File-Tree.html#t:FSTree does not paying attention is path file or directory. It is quckest way to determine this is empty directory on pattern matching level, not using any comparisons.

codeworld-server/src/Util.hs
nixorn: Yes. Storing order in dir.info will resolve this.

web/blocks.html
nixorn: Change what? Insert slash everywhere?

nixorn: Slash is removed.

codeworld-server/src/Main.hs
nixorn: If `dir.info` content from

```
folder_name
```

will turn into something like

```
{
  name:  "folder_name",
  childrenOrder: ["project1", "project2", "project3", "folder1", "folder2"]
}
``` 
It looks inconvenient to use, because I should sort directories/projects not by some field, but sort their in same order as names in list.  

Also projects and folders can have intersecting names, so if one user placed project "123" before or after folder "123" - we should save this position. So ordered list like `["project", "123", "123"]` can not be source of truth.

I going to add integer order field into both `dir.info` and project json, and call `saveProject` handler on user move project.

codeworld-server/src/Util.hs
```diff
@@ -36,8 +37,10 @@ import Data.Monoid
 import Data.Text (Text)
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
+import qualified Data.Text.Lazy as LT
+import qualified Data.Text.Lazy.Encoding as LT
 import System.Directory
-import System.File.Tree (copyTo_, getDirectory)
+import System.File.Tree (Tree(Node), toTree, copyTo_, getDirectory)
```
cdsmith: Here's some bad news.  I'm working on migrating CodeWorld to GHCJS 8.6, and it turns out that I needed to remove uses of `filesystem-trees`.  This package is not maintained, and still depends on `directory` < 1.3, which in turn depends on `base` < 4.11.  That means it cannot be used with modern GHC versions.

We could patch the package locally, but honestly it doesn't do anything really special.  So I'd rather just replace its uses with code that uses `directory` immediately, instead.  I already did this (see the `ghc86` branch) for existing code, but this is adding a new one.

nixorn: Ok, no worries, I'll merge `ghc86` into this branch when work will finished, and will replace it by something more modern.

codeworld-server/src/Util.hs
cdsmith: You're right, you definitely need to load the files in order to determine their names.  That's unfortunate.  But I think we should decide not to keep all of these files in memory, in any case.  Each file can be up to a megabyte or so (and that limit is a bug), and there could easily be a hundred projects in someone's directory tree.  There probably are in mine.

nixorn: "There probably are in mine."

I implemented lazy loading for this tree in last commit.

Now I going to implement route with updating of metadata, to set order of files in directory.

codeworld-server/src/Util.hs
nixorn: I removed `filesystem-trees` from my logic, it nt needed anymore.

cdsmith:
@nixorn I've mostly lost track of the status of this.  Let me know when you're ready for me to review again, and I'll take a fresh look.

One thing worth mentioning is that your editing of the earlier migration script is scary!  I'd rather not do that if we can avoid it.  With multiple migration steps combined in one bit of code, it's hard to reason about or test that it still works for old data formats.  If we must do another migration, it would be better to add a separate "migration step 2" binary, and run both of them from `run.sh`.

But if possible, I'd like to lazily create the order files instead.  If there is no file giving an order, we can assume the files should display in alphabetical order (that's the current behavior)

nixorn:
" Let me know when you're ready for me to review again" - Ok.

Hm, yes, I realized that solution with order files is much better. Because of

1. we do not need other migrations, which is source of production breaks
2. we need access to single file to read or update order, not need to deserialize, update `order` field, and serialize each `dir.info` or `*.cw` in given directory.

I going to change logic one more time, sorry for all this mess.

nixorn:
@cdsmith now there is 2 known bugs

first is not working `Run`. But I think it is not related to this code, because http://127.0.0.1:8080/runBaseJS returns 500 error `getFileStatus: does not exist (No such file or directory)`. I guess its is not working ghcjs installation.

second, which I can not catch, and will try to reproduce as minimal working example and report as jqTree bug
Peek 2019-06-13 14-22

Rest is ready for review.

cdsmith:
I agree that the 500 from runBaseJS isn't related to your code.  I'm not sure why it's happening, but I've changed a lot lately with the GHCJS 8.6 changes, so you may need to delete your build directory and rerun `install.sh` to get that working again.

nixorn:
Issue with quasi-opened folder is fixed.

cdsmith:
This basically looks fine.  It's a huge change and hard to review all at once, so I'll just merge into a test server and do some testing, then merge into master.  Any reason I should hold off?

nixorn:
No reasons!

PR #961, opened 2019-06-03
Two new requirements rules - line length and no warnings

starsandspirals:
This adds two new types of requirement rule to the requirements checker.

The first is line length, which simply takes an integer and fails if any of the lines in the source code are above the length given.

The second is checking for warnings. This can either be strict, in which case it will fail if any warnings arise once the code is compiled, or otherwise a regular expression can be given to provide exceptions to the rule, so it will check for any warnings that do not match the regular expression.

### DIFF
--- a/RequirementChecker.md
+++ b/RequirementChecker.md
@@ -219,6 +219,26 @@ of conditions. Here are the current checks implemented.
 
   Checks that there are no references to `var` in the module.
 
+- `lineLength`
+
+  Example:
+
+      lineLength: 80
+  
+  Checks that there are no lines longer than 80 characters.
+
+-  `noWarnings`
+
+  Example:
+
+      noWarnings:
+        strict: false
+        exceptions: Defined but not used
+
+  If `strict` is true, checks that there are no warnings in the module.
+  If `strict` is false, checks that there are no warnings except for those 
+  matching a regular expression given by `exceptions`.
+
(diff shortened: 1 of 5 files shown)


starsandspirals:
Looks like I was logged into the wrong Google account when I signed the CLA! Should be fixed now.

RequirementChecker.md
```diff
@@ -219,6 +219,26 @@ of conditions. Here are the current checks implemented.
...
+
+  Example:
+
+      noWarnings:
+        strict: false
+        exceptions: Defined but not used
+
+  If `strict` is true, checks that there are no warnings in the module.
+  If `strict` is false, checks that there are no warnings except for those 
+  matching a regular expression given by `exceptions`.
```
cdsmith: Do you really need a boolean flag in addition to the exceptions field?  Wouldn't it be equivalent to just not specify any exceptions?

Why is `exceptions` a singular field?  It looks like a repeated field is better suited.

RequirementChecker.md
```diff
@@ -219,6 +219,26 @@ of conditions. Here are the current checks implemented.
 
   Checks that there are no references to `var` in the module.
 
+- `lineLength`
```
cdsmith: [nit] Can you call it `maxLineLength`?  I think that makes the meaning clearer.

PR #962, opened 2019-06-04
Make local auth working after page reload.

nixorn:


cdsmith:
I'm really not sure that leaving the user logged in forever is a good idea.  Is this necessary for something to work, or just a matter of convenience?

nixorn:
It is very inconvenient to sign in after each page reload(using local auth). I have done it hundreds of times while have worked on projects tree tool. It is waste of time.

Also without this sharing don't work with local auth.

cdsmith:
Filed #967 for the folder sharing issue.  That should be fixed regardless of whether you're on local auth or oauth.  It's also a problem for users who are using oauth, but have just chosen to log out.

For the rest, I think I'm okay with this, but I'll probably add an expiration policy of about a week, so that logins don't last forever,

PR #974, opened 2019-06-10
Two more requirements checker rules - type signatures and whitelisting/blacklisting

starsandspirals:
This adds another two new rule types to the requirements checker.

The first checks that all top-level declarations have accompanying type signatures. The second allows a blacklist or whitelist of symbols to be provided, and in the blacklist case checks that none of the given symbols appear in the code; the whitelist case checks that only the symbols given appear in the code, and no others.

The type signature rule has a boolean that isn't really necessary, as if it were false you just wouldn't include the rule, but I can't find an easy way of having a rule without any value alongside it; if there are any suggestions for this or other adjustments to these rules let me know!


codeworld-compiler/src/CodeWorld/Compile/Requirements/Eval.hs
```diff
@@ -236,6 +277,24 @@ allDefinitionsOf a m = everything (++) (mkQ [] funcDefs) m ++
         patDefs (PatBind _ pat rhs _) | patDefines pat a = [rhs]
         patDefs _ = []
 
+topLevelNames :: Module SrcSpanInfo -> [String]
+topLevelNames m = everything (++) (mkQ [] names) m
```
cdsmith: I don't think this function does what you think it does.  `everything` recursively looks through the entire syntax tree, which means this will also find declarations in `let` and `where` blocks and such.

I actually think that here (and also when looks for type declarations), you don't want to use syb at all!  Instead, just pattern-match on the module constructor and look through that specific list of declarations.  Those are the top-level ones.

codeworld-compiler/src/CodeWorld/Compile/Requirements/Eval.hs
```diff
@@ -223,6 +224,46 @@ checkRule (NoWarningsExcept ex) = do
...
+
+        nameString :: Name SrcSpanInfo -> String
+        nameString (Ident _ s) = s
+        nameString (Symbol _ s) = s
+
+    if | null noTypeSig || not b -> success
+       | otherwise -> failure $ "The declaration of `" ++ head noTypeSig
+           ++ "` has no type signature."
+
+checkRule (Blacklist bl) = withParsedCode $ \m -> do
```
cdsmith: This won't look at the scope, so (for example) if you ban use of `red` in the blacklist, this will also catch:

    pic = colored(solidCircle(1), red)
      where red = RGB(1,0,0)

Ideally, that would still be allowed.

That said, though, fixing this will be much easier in a source plugin, where you can run after renaming.  So maybe this is okay as an interim step.

codeworld-compiler/src/CodeWorld/Compile/Requirements/Eval.hs
starsandspirals: I think I've fixed this!

codeworld-compiler/src/CodeWorld/Compile/Requirements/Eval.hs
starsandspirals: I've had a look at trying to fix this, but I think you're right and it'll be easier to do later on. I've left a note about it in RequirementsChecker.md so we don't forget.

PR #987, opened 2019-06-17
Migrating requirements checker to GHC parser

starsandspirals:
This isn't quite ready to be merged yet, but I thought it was worth opening the pull request in case there's anything that needs to be fixed. I've tested all the rules I've migrated with a few examples, and the compiler unit tests still pass, but they should probably be tested for edge cases and so on.

So far I've migrated the majority of the rule types to use the new GHC parsing as opposed to haskell-src-exts. Things that still need doing are:

- The template matching rule
- The warning checker rule - the CodeWorld diagnostic type uses haskell-src-exts at the moment, so that'll need to be changed along with this
- Once these two rules are migrated, remove the dependency on haskell-src-exts so the qualified imports of ghc-lib-parser aren't needed any more, which should make the code much neater


cdsmith:
By the way, I don't think the unit tests currently exercise the requirements checker at all!  If you've got test cases that you are trying, feel free to add unit tests for them.  It's pretty easy!  You just add a directory under `codeworld-compiler/test/testcase`, and drop a `source.hs` and `expected_output.txt` into it.  It may also be worth creating a new directory for requirements-checker tests, just to keep them separate, and modifying the tests to look in both directories.  Your requirements tests will come in handy later when migrating to a plugin.

cdsmith:
I looked over the code, and although I didn't review in detail and don't know the nuances of the GHC parse tree, it looks like the right approach to me.

If the `GHCParse.` qualifier is a pain, you could pick something shorter, like `GHC.`.  You may not strictly need the package imports, either.  I used them because GHC's package names are so cryptic.  At least this gives an unsuspecting reader some guidance about where to look for what "HsSyn" and the like are about.

cdsmith:
One more amendment to your steps above: haskell-src-exts is used not just in the requirements checker, but also for some other rules I enforce on the educational language variant.  I'll need to migrate those over before the haskell-src-exts dependency can be entirely eliminated.  That's not part of your project, so if you're tolerant of having both, I'd suggest you leave it that way, and continue with your own work.  I'll do the rest of the cleanup later.

starsandspirals:
A couple of updates:

 - I have a working version of the warnings checker rule using the GHC version of warnings and errors, but since at the moment it runs before renaming or typechecking it isn't very useful, as it will only spot parse errors. I think it'll be better to leave this out for now since we'll be staying dependent on haskell-src-exts anyway for other things, and put it back in when we move to a plugin, so it can catch all the same warnings the current version does.

 - I'm having some issues with the template matcher. I'm consistently getting a parse error along the lines of "Perhaps you intended to use TemplateHaskell" from ghc-lib-parser, even though the TemplateHaskell flag is being passed in; the strange part is that I also get a parse error upon trying to use the original unchanged version of the template matcher with the old haskell-src-exts parsing, which I can only assume is the same error. This happens with several of the examples of how to use this rule from RequirementsChecker.md. As far as I can tell I haven't changed anything that should affect this. Any ideas?

cdsmith:
Okay, maybe I broke the template parser when I added the new GHC parsing?  I'll look.

cdsmith:
Yes, it looks like the flags issue is my fault.  I accidentally switched from appending the custom flags onto the end of the list, to appending them on the beginning.  Since the default codeworld-mode flags include `NoTemplateHaskell`, that's overriding the one passed in.  I'll fix it.

cdsmith:
I am hopeful that https://github.com/google/codeworld/commit/d5f647a158029194e093a71ffc1dec8a770aca6e will fix your Template Haskell problem.

chessai:
I'm also curious/hopeful to see if `d5f647a` fixes your problem (what a subtle bug, if that was the fix!). I just took a quick look over this code, and your work looks good. I'm excited for the plugin stage.

starsandspirals:
I've migrated the template matcher code, and for basic cases using things like $any and $var it seems to work fine. However, in more complex cases using things like $contains I'm having another parsing issue that I can't find the cause of; it seems that the GHC parser is failing to parse things containing quotation brackets. An example is below:

`{-
REQUIRES

Description: Contains template test
Rules:
  - containsMatch:
      template: |
        f $var = $(contains [| translated |])
    explanation: f should use translation in its definition.
-}

f(x) = translated(x, 1, 1)
newLogo = f(codeWorldLogo)
program = drawingOf(newLogo)`

This program compiles fine using the old haskell-src-exts parsing, but now when the template is parsed this returns GHCNoParse, so I can't test if the matcher is working or not. I've pushed the code as it stands anyway, and I'll keep looking for a solution to this unless you can spot something that would be causing it!

starsandspirals:
(Oh, and the previous issue was indeed fixed by swapping the order of the flags! Thanks for sorting that one out.)

cdsmith:
This is just a random guess, but have you tried passing `[ "TemplateHaskell", "TemplateHaskellQuotes" ]` for the `extraExts` argument?  Usually, the second extension is implied by the first, but maybe we're working at too low a level in the compiler, and it's not enforcing the implication?

chessai:
@cdsmith if that works, I would consider it a bug. No harm in trying it though.

cdsmith:
The other thing I hope you've tried already is to temporarily modify the GHC-based parsing code to print the error message at https://github.com/google/codeworld/blob/master/codeworld-compiler/src/CodeWorld/Compile/Framework.hs#L216, so that you can see what GHC says is wrong.  If you've tried this, what does it say?

cdsmith:
Something like:

    case GHCParse.unP GHCParse.parseModule state of
        GHCParse.POk _ (GHCParse.L _ mod) -> return (GHCParsed mod)
        GHCParse.PFailed state            -> do
            liftIO $ GHCParse.printSDoc $ GHCParse.ppr $ GHCParse.getMessages state dflagsWithPragmas
            return GHCNoParse

starsandspirals:
Sorry that you went to extra effort, but adding TemplateHaskellQuotes did fix the problem - @chessai I don't know enough to say if this is a bug in ghc-lib-parser or not! 

This allowed me to test the more complex cases, and although it turned out they didn't always work I've fixed some bugs and now the rule works in all the cases from RequirementsChecker.md at least, and a few other examples that I tried. I've pushed the fixed code, but definitely let me know if you find a case where it doesn't behave correctly.

cdsmith:
Is this ready to merge, then?

I think we're happy to take some risks early on with the requirements checker, since I don't believe anyone is using it.  So if you honestly believe it works, then let's merge and move forward.  Any bugs can be fixed later.

It would be good to get some test cases in place soon, though, to verify that future changes don't break anything!  That's probably a good next step.

chessai:
Can we add the test cases before merging, or would you prefer to merge now and work on that later, @cdsmith?

chessai:
I think I'd prefer adding tests now, unless you think otherwise.

starsandspirals:
I'm happy to put together some test cases tomorrow (it shouldn't take long, unless it turns up some bugs in the matcher), so merging now or after that is fine with me either way.

cdsmith:
Okay, test cases now is fine.  I just don't want to build up a bunch of changes without merging.

starsandspirals:
I've written a set of test cases. They don't cover every rule or every edge case, but it hopefully covers most of the bases so that if something breaks we'll know about it. We can add more later if necessary!

cdsmith:
Great!  When you feel this is ready to merge, can you rename to remove the [WIP] label?

starsandspirals:
Yes, sorry! I think it is ready; I don't know for sure all the rules work in every case but they all at least work in the basic cases I've tried, and hopefully the further work in moving to a plugin and creating examples will turn up any bugs.

PR #1033, opened 2019-06-30
Update Blocks UI to API changes

### LINKED ISSUE #672: Blocks UI has fallen behind recent API changes
I need to re-evaluate the block tray, and make sure it's caught up with recent changes like `path` to `polyline`, `activityOf`, and `text` to `lettering`.

ZavierHenry:
Closes #672. This PR should cover all the API changes for the Blocks UI, except for sketches

NB: Considering the other top blocks have an icon associated with them, the same should probably occur with "activityOf". However, I wasn't sure of an icon to use, so I left it blank


cdsmith:
Wow, thanks!

PR #1059, opened 2019-07-15
Bring GHC plugin version of requirements checker back into CodeWorld

starsandspirals:
This pull request adds the codeworld-requirements folder, which is the new source plugin version of the requirements checker. 

Currently this isn't integrated and so the old version is still being used as of this pull request - as Chris suggested, I'll do this work next in a separate pull request.

So far, this handles all the same rules that are currently handled by the CodeWorld requirements checker, except for the no warnings rule. The requirements checker output from this is currently just output as a message at compile time, but I'll get this into the right place when I integrate everything else.

(I wasn't too sure how to handle licensing, but I thought I'd just make the pull request leaving everything as Cabal gave it to me and you can advise me if I need to fix anything!)


cdsmith:
@chessai Can you take a look as well?

codeworld-requirements/src/Framework.hs
```diff
@@ -0,0 +1,129 @@
...
+applyExtensionToFlags dflags name
+  | "No" `isPrefixOf` name =
+        GHCParse.xopt_unset dflags $ fromJust $ M.lookup (drop 2 name) ghcExtensionsByName
+  | otherwise =
+        GHCParse.xopt_set dflags $ fromJust $ M.lookup name ghcExtensionsByName
+
+ghcParseCode :: [String] -> Text -> GHCParsedCode
+ghcParseCode exts src = do
+    let buffer = GHCParse.stringToStringBuffer (T.unpack src)
+        defaultFlags = GHCParse.defaultDynFlags fakeSettings fakeLlvmConfig
```
cdsmith: I think you want to start with the DynFlags that were passed in, right?  It doesn't seem right to use a fake config when there's a real one.

starsandspirals: Yes, I think you're right on this. The only place this is currently used is by the template matcher for parsing the templates, so no situation came up in testing where I noticed it was still using the fake ones! But it'll probably end up getting used for a couple more things where it will matter, so I'll fix it, although it'll involve passing the DynFlags around to a few functions.

chessai: Seconded (thirded?). A lot of subtle bugs can arise from using fake `DynFlags`.

chessai:
This looks good to me on a quick look. I may need to look more after work.

codeworld-requirements/src/Framework.hs
cdsmith: Thinking about this more... there's some subtlety here.  The global DynFlags contains things from LANGUAGE and OPTIONS_GHC pragmas in the current module, right?  That means it's possible for student source code to change the interpretation of the requirements, and possibly cause them to fail to be parsed?

This isn't a new problem, though.  If you agree this is an issue, maybe create an issue to track it, but don't block this pull request on fixing it.  It would even be fine if it didn't get fixed this summer.  I just want to keep track of the known limitation.

codeworld-requirements/src/RequirementsChecker.hs
```diff
@@ -0,0 +1,37 @@
+module RequirementsChecker (plugin) where
```
cdsmith: Google's lawyers require that new source code has an appropriate copyright header, which you can copy from any other file in the project.

codeworld-requirements/codeworld-requirements.cabal
```diff
@@ -0,0 +1,50 @@
+-- Initial requirements-checker.cabal generated by cabal init.  For further
+--  documentation, see http://haskell.org/cabal/users-guide/
+
+name:                codeworld-requirements
+version:             0.1.0.0
+-- synopsis:
+-- description:
+license:             Apache-2.0
+license-file:        LICENSE
+author:              Daniel Marshall
```
cdsmith: There are also legal constraints here.  You should identify the "author" as "The CodeWorld Authors", and then add yourself to the AUTHORS and CONTRIBUTORS files in the root dir of the project.  Again, you should be able to follow the lead of any other cabal file in the project.

codeworld-requirements/src/Framework.hs
chessai: You could just override those, no? Wouldnt setting those to nothing be preferred?

codeworld-requirements/src/RequirementsChecker.hs
starsandspirals: I've now fixed these legal/copyright issues.

codeworld-requirements/src/Framework.hs
starsandspirals: I don't think just categorically overriding them would work - theoretically we want this to be usable for code that makes use of language extensions etc, and it'd be more difficult to tell the difference between ones that are being used for good reasons and ones that will break the requirements. Maybe it could be a temporary solution, though! I'll make an issue as suggested.

cdsmith:
Thanks!  I'll go ahead and merge this.

Issue #1062, opened 2019-07-24
Animations don't work correctly when compiled with ghc (blank-canvas backend)


jhrcek:
**Shared Program**: https://code.world/haskell#PrmAjtMouNGuvXWPJVDeFuQ

**Description**
The simple program above works fine when executed on codeworld.
But when I run it locally and compile it with ghc (with blank canvas) it seems like each frame of the animation is retained, eventually cluttering the screen.

I put together a small project on my fork (see the last commit [here](https://github.com/jhrcek/codeworld/commit/d9db77b47e86cd0fefa747f2e46d73eeb5f5b5b2) to demonstrate the issue).
Steps I use to reproduce the issue:

clone my fork and checkout the branch with the reproducer, build and run with cabal
```
git clone git@github.com:jhrcek/codeworld.git
cd codeworld
git checkout blankCanvasBroken
cabal new-build local-test
cabal new-run local-test
-- open http://127.0.0.1:3000/ in the browser
```
This is how it ends up looking.

![Screenshot from 2019-07-24 07-24-56](https://user-images.githubusercontent.com/2716069/61767425-7fd00a00-ade4-11e9-8441-5a9ca966b9c7.png)

I didn't look into how the blank canvas backend is implemented, but intuitively it seems it should be possible to just erase the canvas before rendering next screen.

I'm on Linux, Fedora 29, building project using cabal-install version 2.4.1.0. and I can reproduce the issue both with google-chrome version 75,  Firefox version 68.

My motivation for using ghc and compiling projects locally: I'd very much like to play with the codeworld api locally using ghcid for instantaneous feedback during development.

jhrcek:
Resolved via https://github.com/google/codeworld/commit/ae2f728da792f9a1b2b8cc8ad02e15c2f20778b7 and https://github.com/google/codeworld/commit/f8d385f5551dc2fb36f61a84bbf40c6db99a6e29. Thanks!

PR #1063, opened 2019-07-24
Clear screen before drawing new Drawing in blank-canvas backend

jhrcek:
This fixes the issue https://github.com/google/codeworld/issues/1062 for me locally. I'm not 100% sure if this is the right place to apply the fix. But this was the main difference I found between ghcjs-based impl (it's calling clear screen before drawing each drawing) and blank-canvas based impl, which wasn't calling it anywhere..


cdsmith:
Thanks for catching this!  I'd mildly prefer to change the `drawDrawing initialDS` to `drawFrame` on the lines right after where you inserted the `clearScreen` calls.  This keeps the code a little closer to the GHCJS variant.

cdsmith:
I'll merge this and tweak afterward.

jhrcek:
Thanks!

PR #1067, opened 2019-07-26
Renaming/typechecking stage for requirements checker

starsandspirals:
This still isn't integrated properly and used by CodeWorld for the requirements checking due to my issues with ghcjs-boot preventing me from building CodeWorld currently, but I thought it was worth making a pull request for the work I've been doing on the requirements checker so I don't get too many commits ahead.

This adds the rules that are checked after renaming or typechecking to the requirements checker, currently including the no warnings rule which is checked after typechecking to ensure that all warnings are counted, and the blacklist/whitelist rules which are now checked after renaming rather than after parsing. (This also means that all the rules which were implemented in CodeWorld are now implemented in the plugin version of the requirements checker.) This pull request also reorganises the structure of modules within the requirements checker, to better match the other CodeWorld projects and in preparation for integration.

### DIFF
--- a/codeworld-requirements/src/CodeWorld/Requirements/Checker/Types.hs
+++ b/codeworld-requirements/src/CodeWorld/Requirements/Checker/Types.hs
@@ -66,11 +66,11 @@ module CodeWorld.Requirements.Checker.Types where
     getStage (NotUsed _) = Parse
     getStage (ContainsMatch{}) = Parse
     getStage (MatchesRegex{}) = Source
-    getStage (OnFailure _ r) = getStage r
+    getStage (OnFailure _ _) = Multiple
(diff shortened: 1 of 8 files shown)


codeworld-requirements/src/CodeWorld/Requirements/Checker/Types.hs
```diff
@@ -66,16 +66,16 @@ module Requirements.Types where
     getStage (NotUsed _) = Parse
```
cdsmith: I don't think you should do it in this pull request, but it seems worth reevaluating whether some of these could be moved to a later stage.  NotUsed seems like it would be more correct during renaming, for example.

Issue #1073, opened 2019-08-07
Imprecise detection of template haskell


cdsmith:
**Shared Program**: https://code.world/haskell#PvmcFl0m1uGPxR-OH8_lVeQ

**Description**

The detection of template-haskell currently just looks for any of "TemplateHaskell", "QuasiQuote" or "glasgow-exts" as a substring of the source file.  This isn't saying exactly what we want.  In particular:

1. This might not be sufficient in the future to detect all ways to turn on TemplateHaskell.  GHC could add new flag sets that have this effect (in the same way that -fglasgow-exts does now).

2. These strings can sometimes occur in comments, string literals, etc., where they are innocent, but the program is rejected anyway.

Copied from a comment on #1072:

> codeworld-compiler now links ghc-lib-parser, and parseDynamicFilePragma can be used to parse language and options pragmas from the source code into dflags, and then we can check for use of the language flags in the resulting DynFlags value. There's existing code for this in CodeWorld.Compile.Framework ghcParseCode.

@misterbeebee FYI

PR #1074, opened 2019-08-08
WIP: Integration of requirements checker plugin

starsandspirals:
I don't know if you're looking at this at the same time - no problem if you are - but I thought it was worth making the current state of my efforts more visible in case it makes any issues become apparent.

I've been following along with the advice received from Luite to try and fix my Cabal/GHCJS issues, but have generally just ended up running into different problems, so in the current state this is stuck at another cabal error.

The current issue is that the package ghc-api-ghcjs cannot be found; I've discovered that it's in the build/ghcjs/lib folder, but I don't know how to access it or other ghcjs packages from there in the cabal file! But as a summary of progress made in this pull request so far, I've gotten rid of the old requirements checker code in the compiler folder that's all duplicated in the new requirements folder, added the required plugin arguments to the call to ghcjs in Compile.hs, and implemented the feedback from Luite, i.e. created a stub version of the plugin and separated out the ghc and ghcjs versions to try and avoid the previous issues.


codeworld-compiler/codeworld-compiler.cabal
```diff
@@ -54,7 +48,6 @@ Library
     directory,
     exceptions,
     filepath,
-    ghc-lib-parser < 8.8,
```
cdsmith: Can you leave this and the stuff in CodeWorld.Compile.Framework?  I already have one other place I'd like to use it (#1073).

starsandspirals: One of the reasons I got rid of this as soon as possible was that when I try to build with it as is, I get the following error preventing me from building:

[2 of 4] Compiling CodeWorld.Compile.Framework ( src/CodeWorld/Compile/Framework.hs, dist/build/CodeWorld/Compile/Framework.o )

src/CodeWorld/Compile/Framework.hs:65:1: error:
    Could not find module ‘Platform’
    Perhaps you meant Platform (needs flag -package-key ghc-8.6.5)
   |
65 | import qualified "ghc-lib-parser" Platform as GHCParse
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Do you have any idea what's causing this? If so, I can bring it back now; if not, I'll leave it out until I've finished working on integration and then bring it back for you to look at, as otherwise I can't build to test anything.

starsandspirals:
Installing the packages using cabal as you suggested worked, so I added this to install.sh.

Now the build process completes successfully again, but I get a new and different error when trying to compile programs using CodeWorld. Now it says:

```
The value plugin did not have the type Plugin as required
```

This doesn't really make sense, as the value of plugin in both RequirementsChecker.hs and the stubbed version definitely has type Plugin, so I don't know where this is coming from; but at least it's a different error.

cdsmith:
@luite Please help!

codeworld-compiler/codeworld-compiler.cabal
cdsmith: It looks like the `Platform` module was removed (or flag-protected?) between 0.20190603 and 0.20190703.  You can add a tighter version bound to ghc-lib-parser in the cabal file.  Sorry about that!

luite:
Ah the error messages there aren't yet as good as they should be. And the plugin loading mechanism is rather hacky right now.

Right now what GHCJS does is the following:

1. the plugin name and module is located in the GHCJS package DB. The type of `plugin` should be `ghc:Plugins.Plugin`, where `ghc` is the GHC library from the GHCJS package db.
2. GHCJS tries to find a module with the same name in the package database specified with `-host-package-db`
3. GHCJS checks that the type of the plugin is `ghc-api-ghcjs:Plugins.Plugin`. This should be the same instance of `ghc-api-ghcjs` that GHCJS is built against.

Most likely 2 or 3 go wrong. What's the easiest way for me to reproduce the build?

codeworld-compiler/codeworld-compiler.cabal
starsandspirals: I've fixed this bound and restored the code in Framework.hs.

starsandspirals:
With the extra information about how GHCJS deals with plugins, I think I've figured out how to arrange the cabal file so that it should work. The new setup looks more like this:

```
flag build-for-ghcjs
  Description: Build for GHCJS
  Default:     False
  Manual:      True

library
  if impl(ghcjs)
    hs-source-dirs:    stub-src
    build-depends:     ghc >= 8.6.5,
                       ghc-boot-th
  else
    hs-source-dirs:    src
    other-modules:     CodeWorld.Requirements.Framework
                       CodeWorld.Requirements.Requirements
                       CodeWorld.Requirements.Checker.Eval
                       CodeWorld.Requirements.Checker.Language
                       CodeWorld.Requirements.Checker.Matcher
                       CodeWorld.Requirements.Checker.Types
    if flag(build-plugin-for-ghcjs)
      build-depends:   ghc-api-ghcjs
    else
      build-depends:   ghc >= 8.6.5,
                       ghc-boot-th
```

which matches closely with the example setup in Luite's email, and with the addition of the flag and extra conditional should install the packages with the right dependencies.

However, I'm having some trouble when actually trying to build with this flag. After adding "-f build-plugin-with-ghcjs" to the cabal install command, I get this error:

```
===== RUNNING: cabal_install ./codeworld-server ./codeworld-error-sanitizer ./codeworld-compiler ./codeworld-requirements ./codeworld-game-api ./codeworld-prediction ./codeworld-api ./codeworld-game-server ./codeworld-account ./codeworld-auth -f build-plugin-for-ghcjs
Warning: the --global flag is deprecated -- it is generally considered a bad
idea to install packages into the global store
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: codeworld-requirements-0.1.0.0 (user goal)
[__1] rejecting: codeworld-requirements:-build-plugin-for-ghcjs (constraint
from config file, command line flag, or user target requires opposite flag
selection)
[__1] trying: codeworld-requirements:+build-plugin-for-ghcjs
[__2] unknown package: ghc-api-ghcjs (dependency of codeworld-requirements
+build-plugin-for-ghcjs)
[__2] fail (backjumping, conflict set: codeworld-requirements, ghc-api-ghcjs,
codeworld-requirements:build-plugin-for-ghcjs)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: codeworld-requirements,
codeworld-requirements:build-plugin-for-ghcjs, ghc-api-ghcjs
```

It's as if ghc-api-ghcjs isn't installed, like before, except that now it is installed and works fine when it isn't being used conditionally as a result of this flag.

Any advice?

cdsmith:
Daniel, I'm so sorry.  I've been swamped at work this week, but I'll try to take a look at this tonight or tomorrow.

Any chance you'll be at ICFP next week?

starsandspirals:
It's no problem at all! I haven't made any progress on this either, unfortunately; I'm happy to keep on pushing with it after GSoC officially finishes, though, so we can get it integrated into CodeWorld properly after we get through all the cabal problems.

I will be at ICFP next week; in fact, I'm writing this at the airport! It'd be great to meet you in person at some point, if you'll be around.

cdsmith:
Excellent!  I'm leaving tomorrow night, and arriving Sunday morning.  I look forward to seeing you there.

I am always happy to accept any collaboration on CodeWorld, but I want to be clear that you shouldn't carry forward GSoC obligations past the coding period.  Luite is usually at ICFP, and maybe the three of us can find some time and solve this quickly.  But if not, you should write a GSoC recap focusing on your GHC plugin running on non-GHCJS code, and show some examples of its use.  That will be a successful project.  The GHCJS and CodeWorld integration will still happen, but I will not consider you obligated to continue work on it.

Issue #1178, opened 2019-09-27
install.sh cabal haddock --html needs an output directory


blackgnezdo:
This also seems to be related to cabal-3.

**Description**
```
===== IN codeworld-base, RUNNING: cabal haddock --html
cabal: option `--html' requires an argument DIR

========== Aborting: Command failed. ==========
DIRECTORY: codeworld-base
COMMAND  : cabal haddock --html
EXIT CODE: 1
===============================================
gnezdo@os-ws:~/codeworld$ echo $?
1
```

cdsmith:
Up until now, `cabal` has put this into `dist/doc/html` in the project directory.  There are symlinks set up that expect this.  I suppose we should continue with that path.

I suppose at the same time that these changes are made, we'd also want to advise people that they need to upgrade to cabal 3.0 if they haven't already, since this does NOT look like a backward-compatible change.

cdsmith:
Subsumed into #1181

Carina-Cai:
I have tried `cabal haddock --html -o dist/doc/html` but it does not work. How exactly do we add the directory?

cdsmith:
@Carina-Cai If you just need to add a directory, then `mkdir -p` would be my assumption.

If there's something else wrong, can you provide more details on what is happening?  I don't have time to try myself at the moment.

Carina-Cai:
Hi Smith, thanks so much for your help! It was solved already.

Issue #1176, opened 2019-09-27
install.sh uses unrecognized v2-install option --symlink-bindir


blackgnezdo:
**Description**

It appears install.sh is a bit too trusting to the environment and ends up using some wrong tools from the system despite having just installed the right tool.
```
./install.sh
...
===== RUNNING: ghcup install-cabal
Installing cabal-install-3.0.0.0 into "/home/gnezdo/codeworld/build/.ghcup/bin"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 4071k  100 4071k    0     0  19.5M      0 --:--:-- --:--:-- --:--:-- 19.5M
Successfully installed cabal-install into
  /home/gnezdo/codeworld/build/.ghcup/bin

===== RUNNING: cabal update
Config file path source is default config file.
Config file /home/gnezdo/.cabal/config not found.
Writing default configuration to /home/gnezdo/.cabal/config
Downloading the latest package list from hackage.haskell.org
===== RUNNING: cabal v2-install happy-1.19.9 alex --symlink-bindir=/home/gnezdo/codeworld/build/bin
cabal: unrecognized 'v2-install' option
`--symlink-bindir=/home/gnezdo/codeworld/build/bin'
gnezdo@os-ws:~/codeworld$ which cabal
/usr/bin/cabal
gnezdo@os-ws:~/codeworld$ cabal --version
cabal-install version 1.24.0.1
compiled using version 1.24.1.0 of the Cabal library
gnezdo@os-ws:~/codeworld$ /home/gnezdo/codeworld/build/.ghcup/bin/cabal --version
cabal-install version 3.0.0.0
compiled using version 3.0.0.0 of the Cabal library

gnezdo@os-ws:~/codeworld$ uname -a
Linux os-ws 4.9.0-11-amd64 #1 SMP Debian 4.9.189-3 (2019-09-02) x86_64 GNU/Linux
gnezdo@os-ws:~/codeworld$ cat /etc/issue
Debian GNU/Linux 9 \n \l

```

blackgnezdo:
I'm mistaken, it is picking up the correct cabal binary. It's just the flag which is invalid.

blackgnezdo:
If I understand the intent correctly, cabal-3 can do the same thing with
`--installdir=$BUILD/bin --install-method=symlink`

I can send a PR.

cdsmith:
[...]

PR #1180, opened 2019-09-28
Make install.sh more reproducible

blackgnezdo:
* Freeze versions of hackage, install-cabal, and ghcjs
* Add pkg-config which is reported as missing by cabal-install

This should help (by avoiding cabal-3.0 for now) with #1176, #1178.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -214,9 +214,10 @@ function fix_libexec_binary {
 if [ ! -f $BUILD/progress/ghcjs ]; then
   run .            cabal v2-install happy-1.19.9 alex --symlink-bindir=$BUILD/bin
   run $BUILD       rm -rf ghcjs
-  run $BUILD       git clone --branch ghc-8.6 --single-branch https://github.com/ghcjs/ghcjs.git
-  run $BUILD/ghcjs git submodule update --init --recursive
+  run $BUILD       git clone --recurse-submodules --branch ghc-8.6 --single-branch https://github.com/ghcjs/ghcjs.git
+  # The following two commands can be combined in git 2.13 using --recurse-submodules.
   run $BUILD/ghcjs git checkout e87195eaa2bc7e320e18cf10386802bc90b7c874
+  run $BUILD/ghcjs git submodule update --init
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -192,8 +193,8 @@ if [ ! -f $BUILD/progress/ghc ]; then
 fi
 
 if [ ! -f $BUILD/progress/cabal-install ]; then
-  run .                  ghcup install-cabal
-  run .                  cabal update
+  run .                  ghcup install-cabal 2.4.1.0
+  run .                  cabal update --index-state='2019-09-28T01:46:00Z'
```
cdsmith: You probably also want to remove the `cabal update` in `build.sh`

install.sh
```diff
@@ -215,6 +216,7 @@ if [ ! -f $BUILD/progress/ghcjs ]; then
   run $BUILD       rm -rf ghcjs
   run $BUILD       git clone --branch ghc-8.6 --single-branch https://github.com/ghcjs/ghcjs.git
   run $BUILD/ghcjs git submodule update --init --recursive
+  run $BUILD/ghcjs git checkout e87195eaa2bc7e320e18cf10386802bc90b7c874
```
cdsmith: Is this the right order? I would have thought we want to checkout the right commit first, then run `git submodule update`, so that if ghcjs later changes the submodule version, we wouldn't build an old ghcjs against a newer submodule?

But I don't really understand git.  If you know what you're doing, ignore me.

blackgnezdo: You are right, what I had didn't checkout the submodules. Since the git version may not be recent enough for the nice single command invocation it is a little long-winded.

install.sh
blackgnezdo: Done

blackgnezdo:
I'm leaving the commits as a chain and expect them squashed on merge. Let me know if you prefer a different protocol.

cdsmith:
Thanks!

cdsmith:
Oh, I don't pay much attention to git hostory, to be honest.  I just added a merge commit.

PR #1201, opened 2019-10-15
exclude ghc-lib-parser-0.20191002

peterbecich:
This version does not have modules `HsExtension` and `HsSyn`

https://github.com/peterbecich/codeworld/blob/63185c0bd283d86b3b7585c75fd8d662b7df33d6/codeworld-compiler/src/CodeWorld/Compile/Framework.hs#L59-L60

http://hackage.haskell.org/package/ghc-lib-parser-0.20191002


peterbecich:
Nevermind, already fixed.  I was out-of-date.

PR #1202, opened 2019-10-15
Stack and CI coverage for some modules

peterbecich:
Do you feel like adding Stack configurations to the project, or is this extra maintenance burden?  This PR provides Stack configurations for 

  - codeworld-account
  - codeworld-api
  - codeworld-auth
  - codeworld-compiler
  - codeworld-error-sanitizer
  - codeworld-game-api
  - codeworld-game-server
  - codeworld-prediction
  - codeworld-server
  - a root project which combines these

`stack build` works for all of these; `stack test` works for all but `codeworld-compiler` and the root project.  The passing tests don't seem to depend on `ghcjs` being available on the path (although it is).

I've had no success with compiling `codeworld-base` and `funblocks-client` with Stack.  Maybe if this is resolved: https://github.com/ghcjs/ghcjs/issues/745

`stack exec codeworld-server` seems to work, but I've only looked at the home page.  Again, I don't know if this depends on `ghcjs` on the path.

Intero seems to work with these Stack configurations.

It also has a YAML for Travis-CI.  Tests for `codeworld-compiler` and the root project are disabled.  If `google/codeworld` has Travis configured, it may be necessary to disable this https://github.com/peterbecich/codeworld/blob/13338d676e84eab26a2311891558364b95205a5c/.travis.yml#L55 for one CI build to allow `stack build` enough time to complete and cache without exceeding 50 minutes.

I'm expecting these to pass: https://travis-ci.org/peterbecich/codeworld/builds/598007572


peterbecich:
An easy improvement I could make to this is a common file for `extra-deps`, to avoid duplication.

codeworld-account/stack.yaml.lock
```diff
@@ -0,0 +1,12 @@
+# This file was autogenerated by Stack.
```
cdsmith: This comment suggests that this shouldn't be checked in, if it's autogenerated?  Can we add it to `.gitignore`?  Same for others.

codeworld-api/stack.yaml
```diff
@@ -0,0 +1,18 @@
+resolver: lts-14.9
+
+packages:
+  - .
+
+allow-newer: false
+
+extra-deps:
```
cdsmith: It's unfortunate that we need to list the dependencies twice here.  Is this really necessary?  Stack cannot get this information from the Cabal file instead?  I thought it still used cabal files for dependencies.

stack.yaml
```diff
@@ -0,0 +1,32 @@
+resolver: lts-14.9
```
cdsmith: I don't understand why you need this at the top level AND in the subdirectories.  Isn't that yet another duplication?  Can we just do it at the top level?  That would be far more tolerable.

.travis.yml
```diff
@@ -0,0 +1,55 @@
...
+  - name: "root (no tests)"
+    env: MODULE=. TEST=false
+  - name: "account"
+    env: MODULE=codeworld-account TEST=true
+  - name: "api"
+    env: MODULE=codeworld-api TEST=true
+  - name: "auth"
+    env: MODULE=codeworld-auth TEST=true
+  - name: "compiler (no tests)"
+    env: MODULE=codeworld-compiler TEST=false
```
cdsmith: This is particularly unfortunate, since these tend to be the tests that I care the most about.  But it's not unexpected, since this would require the patched ghcjs built by the install script.

codeworld-account/stack.yaml.lock
peterbecich: Looks like it is optional to check in the `stack.yaml.lock` files: https://docs.haskellstack.org/en/stable/lock_files/#lock-files

codeworld-api/stack.yaml
peterbecich: No, not necessary.  I was using Stack incorrectly.  Only one `stack.yaml` is necessary.

stack.yaml
peterbecich: Yes

codeworld-account/stack.yaml.lock
cdsmith: Okay, I see the point there, but I just don't agree in this context.  Since stack isn't the main build process, I'd rather reduce the maintenance load for it, and these are files that would have to be updated as things change.  So please remove `stack.yaml.lock` from the pull request, and add it to `.gitignore` instead?

cdsmith:
It would be good to add a note to the README clarifying that the `stack.yaml` file is there to support use of intero and travis, and that it cannot be used to build and run CodeWorld.

peterbecich:
Sorry for my late response.

As for the failure of the `codeworld-compiler` tests on Stack, I'm not sure if this is the error you are describing
> because you're getting errors that the modules aren't exporting program?
```
missingEqualSign: [Failed]
missingEqualSign
expected: "program.hs:16:1-9: error:\n    Parse error: An equation was expected here.\n    Are you missing the equal sign?"
 but got: ": Loaded package environment from /tmp/stack19221/test-ghc-env\n<command line>: cannot satisfy -hide-package codeworld-api\n    (use -v for more information) (no loc)"
```

The other failures are all similar to this one.

```
         Test Cases   Total       
 Passed  11           11          
 Failed  33           33          
```

If we can get `stack test codeworld-compiler` to succeed locally I'm sure it could work on Travis in the future, assuming the 50 minute timeout is enough to set up GHCJS.

cdsmith:
This looks valuable as-is.  Thanks!  I'll go ahead and merge it, and any further work on getting more of the project to build can happen in new PRs.

Issue #1229, opened 2019-11-05
Travis build is failing


cdsmith:
The new travis build succeeded a few times, but now it's failing with some kind of message about package corruption (files missing from the `text` package).

cdsmith:
@peterbecich  Any ideas?

cdsmith:
```
/home/travis/build/google/codeworld/codeworld-api/src/CodeWorld/CanvasM.hs:25:1: error:
    Could not find module ‘Data.Text’
    There are files missing in the ‘text-1.2.4.0’ package,
    try running 'ghc-pkg check'.
    Use -v to see a list of the files searched for.
   |
25 | import Data.Text (Text)
   | ^^^^^^^^^^^^^^^^^^^^^^^
```

cdsmith:
That problem went away, but now we've got a timeout running `travis_apt_get_update`.  This seems to be very flaky, and I don't know what to do about it.

cdsmith:
Seems to be okay for now.

peterbecich:
@cdsmith please delete all the Travis caches, I'm guessing here https://travis-ci.org/google/codeworld/caches

cdsmith:
@peterbecich Done

PR #1231, opened 2019-11-07
attempt to make Travis CI reliable

peterbecich:
https://github.com/google/codeworld/issues/1229

I'm not certain this will make Travis CI reliable.  The multiple stages to build dependencies were probably redundant.  I have consolidated those.  Maybe those caused some sort of collision which made package `text` unavailable.

If it persists we can disable caching entirely (50+ minutes per build, then?)  Hopefully we won't have to resort to disabling concurrent builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ cache:
 
 addons:
   apt:
+    update: true
     packages:
       - libgmp-dev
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -32,55 +33,11 @@ jobs:
         - travis_retry curl -L https://get.haskellstack.org/stable/linux-x86_64.tar.gz | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
         - stack setup
     - stage: Build dependencies
-      name: "root"
+      name: "Build dependencies"
       script:
+        # - stack clean --full  # if unreliability persists
```
peterbecich: a.f.a.i.k. dependencies are stored in `~/.stack`, and this does nothing to fix dependency errors

.travis.yml
```diff
@@ -19,6 +19,7 @@ cache:
 
 addons:
   apt:
+    update: true
```
peterbecich: wild guess to prevent `travis_apt_get_update` error (or move the error to the beginning of the build?)

cdsmith:
Thanks.  Worth a try!

PR #1270, opened 2019-12-21
Cabal project file

peterbecich:
`cabal new-build all` works.  

`cabal new-test all` compiles but the `codeworld-compiler` module tests have other problems, unrelated to the `cabal.project` file as far as I can tell.


cdsmith:
Thanks!

PR #1272, opened 2019-12-23
Travis coverage for GHCJS-dependent packages

peterbecich:
Includes https://github.com/google/codeworld/pull/1263


peterbecich:
Also includes https://github.com/google/codeworld/pull/1276

@cdsmith , please review

This PR is one step closer to eventually having all of CodeWorld's tests running in CI: https://github.com/google/codeworld/pull/1202#discussion_r335749998

I assume for some tasks there is no substitute for a GHCJS installation, rather than `ghcjs-base-stub` or `jsaddle`.  Maybe this is not true.

PR #1277, opened 2019-12-28
Stack reinstalls GHC in Travis

peterbecich:
I believe this particular failure https://travis-ci.org/google/codeworld/jobs/630081765#L372-L408 can be prevented by this PR.  `stack setup --reinstall` was already used elsewhere in the Travis config but not in this failing stage.

I can't explain why some things are cached successfully,
```
10.34s$ travis_retry eval $"stack upgrade --binary-only ; sleep 10"
Current Stack version: 2.1.3, available download version: 2.1.3
Skipping binary upgrade, you are already running the most recent version
```
(presumably not Stack from Ubuntu 18.04 image)
and other things are not:
```
Looked for sandboxed compiler named one of: ["ghc-8.6.5","ghc"]
Could not find it on the paths ["/home/travis/.stack/programs/x86_64-linux/ghc-8.6.5/bin/"]
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -134,6 +134,7 @@ jobs:
         - export PATH=$HOME/.local/bin:$PATH
         - travis_retry eval $"stack upgrade --binary-only ; sleep 10"
         - hash -r
+        - travis_retry eval $"stack setup --reinstall ; sleep 10"
         - |
           stack build \
             base-compat \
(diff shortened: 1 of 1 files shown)


peterbecich:
I don't think Travis caches nested directories very well, which reminds me why all of these `ghcjs` dirs are listed explicitly:
https://github.com/google/codeworld/blob/deeb682f11e9ce49c464a8b4b827c9b418cb52c5/.travis.yml#L8-L24

PR #1292, opened 2020-01-09
use Applicative not Monad for Dynamic

### LINKED ISSUE #1057: Reflex programs are very slow
**Shared Program**: https://code.world/haskell#P6gA0a82KeXSlz2_HBxOGxA

**Description**

This is a trivially simple program.  It just spins a single picture on the screen.  But it runs roughly, going through cycles of poor frame rate about once per second.  The slowness seems to be independent of being in debug mode, or of the screen size, so it looks like it's Reflex or my integration that's producing garbage rapidly.

xplat:
should fix #1057

### DIFF
--- a/codeworld-api/src/CodeWorld/Driver.hs
+++ b/codeworld-api/src/CodeWorld/Driver.hs
@@ -10,6 +10,7 @@
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE JavaScriptFFI #-}
+{-# LANGUAGE LambdaCase #-}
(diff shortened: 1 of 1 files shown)


cdsmith:
@xplat 

Thanks for the pull request and the debugging.  I definitely imagined that something much smarter was going on with managing subscriptions here.

It looks like you need to do some paperwork before I can accept this pull request.  If you don't want to do this, then I can take the results of your debugging and reimplement the fix myself.  Let me know what your preference is.

cdsmith:
Thanks for the debugging.  I have implemented something similar myself (CLA/paperwork reasons), and pushed the change.  It definitely helps!  However, the monadic API is still noticeable more jerky than the original one.  So this seems to have been part of the answer, but not the entire answer.

PR #1304, opened 2020-02-08
Added JS logic for autosave and autosave button

### LINKED ISSUE #643: Auto-save
Also, make it really obvious when you haven't saved yet.

devanshbatra04:
closes #643 

I am able to turn on auto save by clicking a button in the Buttons toolbar. How do we notify the user regarding the save? I am planning to add a human readable timestamp of the save somewhere on the GUI.


devanshbatra04:
ezgif com-video-to-gif
Here's the progress so far. I think I should add the save timestamp to the server somehow, so the user knows the last save of the program when it's loaded first time and autosave is not yet on. Something feels missing.

cdsmith:
I'm so sorry I missed this.  Looking now.

cdsmith:
Low-level comments inline, some trivial and some rather deep.

At a high level, though, I'm concerned that adding auto-save as a toggle button adds UI complexity, where my hope was to reduce it.  For example, it doesn't at all resolve the question of how to coordinate saves between collaborative editing among multiple users, because they may not even have auto-save enabled.

I think we want to do something like this:

- Remove the Save button entirely.  The Save As button should remain as it is.
- If you are logged in and working on a saved project, then auto-save is always on.

devanshbatra04:
> Low-level comments inline, some trivial and some rather deep.

Sorry, I didn't understand that. 

> it doesn't at all resolve the question of how to coordinate saves between collaborative editing among multiple users, because they may not even have auto-save enabled.

Understood, I will close this PR and open a new one instead.

Issue #1301, opened 2020-01-31
Install script failing, update cabal index-state


devanshbatra04:
**Description**
The install.sh script is failing. I have tried it on multiple systems (clean ubuntu 18 and debian - both with and without prior Haskell and cabal environments). 

This is what the output looks like.
```
===== RUNNING: cabal v2-install happy-1.19.9 alex --symlink-bindir=/home/jupyter/codeworld/build/bin
cabal: Could not resolve dependencies:
[__0] trying: codeworld-server-0.1 (user goal)
[__1] next goal: ormolu (dependency of codeworld-server)
[__1] rejecting: ormolu-0.0.1.0 (conflict: codeworld-server => ormolu>=0.0.2)
[__1] fail (backjumping, conflict set: codeworld-server, ormolu)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: codeworld-server, ormolu
```

Interestingly, this can be resolved by removing the index-state parameter at in [install.sh](https://github.com/google/codeworld/blob/master/install.sh#L197) (or replacing the value with the current timestamp).

My thoughts are that hackage didn't have the required version of ormolu at the timestamp mentioned in install.sh and it wasn't spotted because Travis builds have been failing (never reaching the build phase). I haven't sent a PR because I am not entirely sure of the matter.

cdsmith:
It sounds like updating the timestamp is the correct solution.  I probably updated the ormolu version and forgot to update the install script.  If you a timestamp that definitely works, please send a pull request or let me know what it is.  Otherwise, I'll test one tonight.

devanshbatra04:
Okay, sure. I will send a PR today. Sorry for delay, had a busy week.

cdsmith:
Looks like Greg's PR fixed this.

PR #1307, opened 2020-02-09
Minor house-keeping to get install a bit further along

blackgnezdo:


blackgnezdo:
#1301 is probably addressed by this change.

blackgnezdo:
The CI failure didn't seem caused by the change (nor is it particularly informative)

PR #1341, opened 2020-04-25
adding filter to codeworld-base

bbarker:
(untested)

### DIFF
--- a/codeworld-base/src/Internal/Prelude.hs
+++ b/codeworld-base/src/Internal/Prelude.hs
@@ -42,6 +42,7 @@ module Internal.Prelude
     -- List functions
     , (P.++)
     , empty
+    , filter
     , contains
     , length
     , at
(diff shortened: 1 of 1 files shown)


cdsmith:
At the very least, this would need to be uncurried to match the standard calling conventions in codeworld-base.

Then, what's the advantage for students to write `filter(p, list)` or `filter(list, p)`, rather than `[ x | x <- list, p(x) ]`?  The latter is very close to already established math notation, and students already learn and use list comprehensions as part of the standard CodeWorld sequence.

cdsmith:
For historical context, `map` and `filter` used to be in the `codeworld-base` prelude.  They were removed because there's nothing you can do here that cannot be done with list comprehensions, too.  Folds are trickier and stuck around much longer, but they didn't really carry their weight either.

cdsmith:
If you want this available for more experimental purposes, feel free to add to a module in the Extras hierarchy.  Something like `Extras.List` perhaps.  `Extras` has a far lower curation barrier than changes to the prelude.

bbarker:
All good points, thanks!

bbarker:
I suppose I did think of one (possible) advantage of having `filter` as predefined a function - you can teach it, in particular to younger students that may have some understanding of functions, without having to first go over the list-comprehension (set-theoretic-esque) notation. I'm not sure how much of an advantage that is yet, as I've not tried to teach list comprehensions yet.

But as you suggested, I'll change the PR to be against `Extras.List`

PR #1360, opened 2020-06-05
Use directory tree state as a single source of truth.

Powell-v2:
This commit eliminates `window.openProjectName` property. Rely on directory tree state to determine the name of the currently open project instead.

### DIFF
--- a/web/js/codeworld.js
+++ b/web/js/codeworld.js
@@ -782,9 +782,9 @@ function loadSample(code) {
 }
 
 function newProject() {
-    utils.directoryTree.clearSelectedNode();
-
     warnIfUnsaved(() => {
+        utils.directoryTree.clearSelectedNode();
+
         setCode('');
     });
 }
--- a/web/js/codeworld.js
+++ b/web/js/codeworld.js
@@ -589,13 +589,13 @@ function setMode(force) {
 }
 
 function getCurrentProject() {
-  const doc = window.codeworldEditor.getDoc()
+    const doc = window.codeworldEditor.getDoc();
 
-  return {
-    name: utils.directoryTree.getCurrentProjectName() || 'Untitled',
-    source: doc.getValue(),
-    history: doc.getHistory()
-  }
+    return {
+        name: utils.directoryTree.getCurrentProjectName() || 'Untitled',
+        source: doc.getValue(),
+        history: doc.getHistory()
+    };
 }
 
 /*
@@ -662,7 +662,7 @@ function updateUI() {
     document.getElementById('saveAsButton').style.display = '';
     document.getElementById('runButtons').style.display = '';
 
-    let title = utils.directoryTree.getCurrentProjectName() || '(new)'
+    let title = utils.directoryTree.getCurrentProjectName() || '(new)';
 
     if (!isEditorClean()) {
         title = `* ${title}`;
@@ -782,11 +782,11 @@ function loadSample(code) {
 }
 
 function newProject() {
-  utils.directoryTree.clearSelectedNode()
+    utils.directoryTree.clearSelectedNode();
 
-  warnIfUnsaved(() => {
-    setCode('')
-  })
+    warnIfUnsaved(() => {
+        setCode('');
+    });
 }
 
 function newFolder() {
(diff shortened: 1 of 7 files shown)


Powell-v2:
@googlebot I signed it!

web/js/codeworld.js
```diff
@@ -585,12 +583,13 @@ function setMode(force) {
 }
 
 function getCurrentProject() {
-    const doc = window.codeworldEditor.getDoc();
-    return {
-        'name': window.openProjectName || 'Untitled',
-        'source': doc.getValue(),
-        'history': doc.getHistory()
-    };
+  const doc = window.codeworldEditor.getDoc()
```
cdsmith, line 586: Let's stick with semicolons after lines.  I believe running format.sh will add them for you.

web/js/codeworld.js
```diff
@@ -784,9 +776,11 @@ function loadSample(code) {
 }
 
 function newProject() {
-    warnIfUnsaved(() => {
-        setCode('');
-    });
+  utils.directoryTree.clearSelectedNode()
```
cdsmith, line 779: Shouldn't this be inside of warnIfUnsaved?  If I cancel the action, I don't think clearSelectedNode should be called.

Powell-v2: Yes, it indeed should be inside of `warnIfUnsaved`.

web/js/codeworld_shared.js
```diff
@@ -982,41 +1004,39 @@ function saveProjectAsBase(successFunc) {
...
+        closeOnConfirm: false,
     }).then(result => {
         const parent = getNearestDirectory_();
 
         function localSuccessFunc() {
-            const matches = parent.children.filter(n => n.name === result.value && n.type === 'project');
+            const matches = parent.children
+                .filter(n => n.name === result.value && n.type === 'project');
             let node;
+            const type = 'project';
```
cdsmith, line 1030: Just nitpicking, but I don't see how these two constants (`type` and `name`) make anything clearer.

web/js/codeworld_shared.js
```diff
@@ -1025,24 +1045,27 @@ function saveProjectAsBase(successFunc) {
...
     });
 }
 
 function saveProjectBase(path, projectName, mode, successFunc) {
     if (!signedIn()) {
         sweetAlert('Oops!', 'You must sign in to save files.', 'error');
-        updateUI();
-        return;
+
+        return updateUI();
```
cdsmith, line 1068: What's going on here?  `updateUI` doesn't have a return value, does it?

web/js/codeworld_shared.js
```diff
@@ -1069,26 +1092,30 @@ function saveProjectBase(path, projectName, mode, successFunc) {
 
         sendHttp('POST', 'saveProject', data, request => {
             sweetAlert.close();
+
             if (request.status !== 200) {
-                sweetAlert('Oops!',
+                return sweetAlert(
```
cdsmith, line 1097: Same question.  I don't see a meaningful return value there.

web/js/codeworld_shared.js
```diff
@@ -1097,55 +1124,61 @@ function saveProjectBase(path, projectName, mode, successFunc) {
...
 
 function deleteProject_(path, buildMode, successFunc) {
-    if (!window.openProjectName) return;
+    if (!utils.directoryTree.getCurrentProjectName()) return;
 
     if (!signedIn()) {
         sweetAlert('Oops', 'You must sign in to delete a project.', 'error');
-        updateUI();
-        return;
+        return updateUI();
```
cdsmith, line 1140: same question about the return

web/js/codeworld_shared.js
```diff
@@ -1210,31 +1247,43 @@ function createFolder(path, buildMode, successFunc) {
             sweetAlert.close();
             const data = new FormData();
             data.append('mode', buildMode);
-            if (path === '') {
-                data.append('path', result.value);
-            } else {
+
+            path === '' ?
```
cdsmith, line 1251: I don't think you should use the ternary operator when you're ignoring the expression's value.  if/else is the right tool here.

web/js/codeworld_shared.js
```diff
@@ -1210,31 +1247,43 @@ function createFolder(path, buildMode, successFunc) {
...
+                data.append('path', result.value) :
                 data.append('path', `${path}/${result.value}`);
-            }
 
             sendHttp('POST', 'createFolder', data, request => {
                 if (request.status !== 200) {
-                    sweetAlert('Oops',
+                    return sweetAlert(
```
cdsmith, line 1257: ditto about the return value.

web/js/codeworld_shared.js
```diff
@@ -1210,31 +1247,43 @@ function createFolder(path, buildMode, successFunc) {
...
+                        'error',
+                    );
                 }
+
                 successFunc();
+
                 let node = $('#directoryTree').tree('getSelectedNode');
-                if (!node) node = $('#directoryTree').tree('getTree');
-                if (node.type !== 'directory') node = node.parent;
+                const type = 'directory';
```
cdsmith, line 1267: As before, I don't think these constants improve readability.

web/js/codeworld_shared.js
```diff
@@ -1360,10 +1409,9 @@ function shareFolder_(mode) {
         updateUI();
         return;
     }
-    if (!getNearestDirectory() || window.openProjectName) {
+    if (!getNearestDirectory() || utils.directoryTree.getCurrentProjectName()) {
         sweetAlert('Oops!', 'You must select a folder to share!', 'error');
-        updateUI();
-        return;
+        return updateUI();
```
cdsmith, line 1414: Ditto with the return value.

web/js/utils/directoryTree.js
```diff
@@ -0,0 +1,38 @@
...
+function clearSelectedNode() {
+    const directoryTree = $(directoryTreeId);
+
+    const state = directoryTree.tree('getState');
+    state.selected_node = [];
+    directoryTree.tree('setState', state);
+}
+
+function createNodeId(type, name) {
+    return `jqtree-element__${type}__${name}`;
```
cdsmith, line 14: One thing to test here is creating two projects in different folders with the same base project name.  Would they both get the same id?

web/js/codeworld.js
```diff
@@ -1314,15 +1325,11 @@ function parseCompileErrors(rawErrors) {
...
-    const treeState = $('#directoryTree').tree('getState');
-    treeState.selected_node = [];
-    $('#directoryTree').tree('setState', treeState);
+    utils.directoryTree.clearSelectedNode();
+
     setCode('');
 }
 
 function clearCode() {
-    window.openProjectName = null;
```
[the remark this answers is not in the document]
Powell-v2: I agree, will take care of it.

web/js/codeworld_shared.js
```diff
@@ -905,23 +915,35 @@ function loadSubTree(node, callback) {
...
+
         data.append('mode', window.projectEnv);
         data.append('path', path);
+
         showLoadingAnimation();
-        sendHttp('POST', 'listFolder', data, request => {
+
+        sendHttp('POST', 'listFolder', data, (request) => {
             if (request.status === 200) {
+                const treeNodes = JSON.parse(request.responseText);
```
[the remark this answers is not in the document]
Powell-v2: Yes, certainly!

web/js/codeworld_shared.js
Powell-v2: Just keeping things DRY. See code right below:

```js
$('#directoryTree').tree(
  'appendNode', {
    id: utils.directoryTree.createNodeId(type, name),
    name,
    type,
    data: JSON.stringify(getCurrentProject()),
  },
  parent
);
```

web/js/codeworld_shared.js
Powell-v2: No, it doesn't have a return value, it just performs side effects. The intention behind this is to reduce the amount of lines as `return` is simply used to interrupt function execution, hence anything returned will be ignored anyway. I realize it can be confusing - should I revert this and other similar changes?

web/js/utils/directoryTree.js
Powell-v2: Good point, there will be collisions, unless we tuck on some unique value. Will something like this be acceptable?

```js
jqtree-element__${type}__${name}__${Math.floor(Math.random() * 1000000)}
```
Or we need to pull in a helper for generating unique IDs?

Powell-v2: You are right, a simple counter should be sufficient here - I was overthinking it.

PR #1369, opened 2020-06-11
Example: Prettier formatter and Standard linter

Powell-v2:


web/blocks.html
```diff
@@ -17,101 +17,173 @@
...
+        </div>
+      </div>
+      <div id="toolbar">
+        <div id="buttons">
+          <a
+            id="navButton"
+            class="cw-button blue"
+            onclick="window.mainLayout.toggle('west')"
+            ><i class="mdi mdi-18px mdi-menu"></i
+          ></a>
```
cdsmith, line 133: What happened to `</i>` here?  Frankly, this is really ugly.  The tag didn't even start at that indent level.  :(  Can it be turned off?

Same below.

third_party/codemirror-buttons/buttons.js
```diff
@@ -1,68 +1,69 @@
 (function (mod) {
```
cdsmith, line 1: We should exclude `third_party` from linting and formatting.  These are just vendored dependencies, and we're not maintaining them except to occasionally import a new upstream version.

web/blockframe.html
```diff
@@ -41,56 +40,56 @@
...
-            frameId.setAttribute('height', maxY);
-        }
-
-        function setId(j_id_param) {
-            frameId = j_id_param;
+      let upperParent = null;
+      let frameId = null;
+
+      function loadXml(data) {
+        let workspace = Blockly.getMainWorkspace();
```
cdsmith, line 54: Hmm, judging by all these `let`s that should be `const`, ESLint seems to be skipping this.  I guess because it's a mixed file type, but that means it's more important to avoid these big blocks of inline JavaScript.

web/js/blocks/cw-lists.js
```diff
@@ -14,835 +14,846 @@
...
+goog.require('Blockly.Blocks')
+
+const listsHUE = 260
+
+Blockly.Blocks.lists_comprehension = {
+  /**
+   * Block for creating a list with any number of elements of any type.
+   * @this Blockly.Block
+   */
+  init () {
```
cdsmith, line 30: What's with these spaces?  It's not even consistent; it's adding a space at definition, but not at function application.  Would definitely prefer to get rid of these.

web/js/blocks/cw-event.js
```diff
@@ -14,61 +14,88 @@
  * limitations under the License.
  */
 
-'use strict';
+'use strict'
```
cdsmith, line 17: I really don't want all the semicolons deleted from JavaScript code.

web/blocks.html
Powell-v2: Wow. That's bizarre. I agree with you that it's really ugly. Should've been:

```html
<a
  id="navButton"
  class="cw-button blue"
  onclick="window.mainLayout.toggle('west')"
>
  <i class="mdi mdi-18px mdi-menu"></i>
</a>
```

I'll look into it.

third_party/codemirror-buttons/buttons.js
Powell-v2: Yes, makes sense. I formatted and linted JS located under `web/js`. I believe there are files symlinked to the `third_party` directory?

web/blockframe.html
Powell-v2: Probably. Indeed, inline JS should be avoided whenever possible.

web/js/blocks/cw-lists.js
Powell-v2: I don't think I understand the reasoning behind this rule either. Good news is that it could be turned off.

web/js/blocks/cw-event.js
Powell-v2: Could you explain why, in a nutshell? I've been writing JS code without semicolons for several years now without running into any issues - although I have to admit that there was always a bundler involved.

third_party/codemirror-buttons/buttons.js
cdsmith: Yes, and that's important.  They have to be located in `third_party` for licensing purposes, but they need to be in `web` for serving.

I've found it useful to do something like `$(find ./web -type f)`, which will list only regular files and not symlinks.

web/js/blocks/cw-event.js
cdsmith: Sure.  It comes down to a few things:

1. Almost all JavaScript I've seen elsewhere uses semicolons, and there's value in doing things the way most others do them.  It makes the code a little easier for new programmers to work with.  I've worked at Facebook and Google, and both of their style guides require them.  I just looked around for polls and surveys and such, and consistently about 90% of JavaScript code and 90% of people report using explicit semicolons.
2. Automatic semicolon insertion is complex, and I believe syntax should be simple and predictable.  I worry about syntax mistakes that are turned into valid code by automatically adding semicolons, and I'd prefer to let ESLint clarify what's going on by adding the semicolon there.
3. Though rare, there have even been cases where changes to JavaScript in browsers have broken older code because new syntax prevented a semicolon from being automatically added.

But, mostly it's the first one.  If most of the JavaScript people see has semicolons, removing them just makes this code base harder to work in.

cdsmith:
So here's a decision: let's keep using ESLint and not Standard.  But Prettier instead of js-beautify sounds good to me.

What I don't like about standard is that its entire value seems to come from making it harder to change the configuration.  I already have an ESLint configuration here, and I'm much happier with that than I am with the one Standard is pushing.  So why not keep that one?

web/blocks.html
cdsmith: Looking at this again, I suspect the reason for that formatting is that adding whitespace between tags is actually a semantic change to the document, while whitespace inside a tag is ignored by the parser.  I've seen this make a difference before.  js-beautify did something similar with string interpolation in JavaScript, changing this:

```
    let a = "this is a really long string " + someExpr()
        + " and some more really really long string";
```

into this:

```
    let a = `this is a really long string ${someExpr()
        } and some more really really long string`;
```

I bet, in this case, if you added a space into the source document, prettier would stop breaking the line inside the tag.

web/blockframe.html
cdsmith: Okay, I'll add that to the code health github project.

web/js/blocks/cw-event.js
Powell-v2: Got it. I don't care too much about this. For what it's worth, my biggest pet peeve is excessive indentation - I'd really like to keep nesting to a minimum.

Powell-v2:
I didn't realize you already have an ESLint config in place. In that case, we should continue using it.

web/blocks.html
Powell-v2: I think you are on point about whitespace sensitivity =>

https://prettier.io/blog/2018/11/07/1.15.0.html#whitespace-sensitive-formatting

Powell-v2:
Prettier team helpfully provides a [bash script](https://prettier.io/docs/en/precommit.html#option-6-bash-script) which could be used in git's pre-commit hook. I think it would be faster if you set things up (running Prettier+ESLint+Ormolu on staged files), but I definitely could take a stab at it as well.

Regarding `format.sh`, I noticed that modules are installed every time this script is executed. Wouldn't it be better to install required tools only once during install / build stage? Or avoid installation altogether by leveraging [`npx`](https://blog.npmjs.org/post/162869356040/introducing-npx-an-npm-package-runner).

cdsmith:
Yes, I definitely support formatting in a git precommit hook, as well as avoiding the constant reinstall.  I think I was just lazy when I set this up initially.  `npx` looks like a perfect answer, but I just wasn't aware of it.

cdsmith:
I'm also fine with setting the indent level to 2 instead of 4.  Especially since I've already switched the Haskell code to Ormolu, which requires a 2-space indent, so it's really more consistent anyway.

web/blocks.html
cdsmith: So this isn't a big deal.  When the whitespace doesn't hurt, we can add it by hand.

cdsmith:
So my understanding of the plan is this:

1. Use prettier to replace js-beautify.
2. Change the ESLint configuration to 2-space indents rather than 4.
3. Modify the formatting script to use `npx` rather than installing new packages each time.
4. Set up a git precommit hook to run format.sh on every commit.

These sound like great changes.

Powell-v2:
Yep, this is exactly what I had in mind.

Issue #1375, opened 2020-06-15
Improve formatting and linting


cdsmith:
@Powell-v2 is working on this.  The plan:

1. Use Prettier to replace js-beautify.
2. Change the ESLint configuration to 2-space indents rather than 4.
3. Modify `format.sh` to use `npx` rather than installing new packages each time.
4. Set up a git precommit hook to run `format.sh` on every commit, on just the modified files.

Powell-v2:
Link to the PR => #1378

PR #1378, opened 2020-06-16
Set up formatting&linting in pre-commit hook (#1375)

Powell-v2:
I ended up using the [script](https://prettier.io/docs/en/precommit.html#option-6-bash-script) provided by the Prettier team.

It's worth noting that I removed the following part:

`[ -z "$FILES" ] && exit 0`

It caused the script to terminate early without actually doing anything useful. What is the purpose of this statement? Is it important to have it?

Tested on JS, CSS and HTML files. Please verify with Haskell files. Please also let me know if something could be improved as I'm not very fluent in bash scripting.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -11,7 +11,7 @@ parserOptions:
 rules:
   indent:
     - error
-    - 4
+    - 2
   linebreak-style:
     - error
     - unix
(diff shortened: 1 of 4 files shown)


cdsmith:
Merged, and I also ran the formatter on everything.  You'll want to pull before doing anything else.

PR #1386, opened 2020-06-20
Convert core JS files into modules

Powell-v2:
I aimed to just make things work again after moving methods around and getting rid of inlined click handlers, without attempting to introduce any extra changes. Both main editor and blocks appear to be functional for me, although I didn't spend a lot of time testing.

### DIFF
--- a/web/blockframe.html
+++ b/web/blockframe.html
@@ -79,13 +79,13 @@
       document.getElementById("blocklyDiv").addEventListener(
         "click",
         () => {
-          if (upperParent) {
+          if (window.top) {
             let workspace = Blockly.getMainWorkspace();
 
             let xml = Blockly.Xml.workspaceToDom(workspace);
             let xml_text = Blockly.Xml.domToText(xml);
 
-            upperParent.loadSample(xml_text);
+            window.top.postMessage({ type: "loadSample", code: xml_text }, "*");
           }
         },
         false
--- a/web/blockframe.html
+++ b/web/blockframe.html
@@ -79,13 +79,16 @@
       document.getElementById("blocklyDiv").addEventListener(
         "click",
         () => {
-          if (window.top) {
+          if (window.parent) {
             let workspace = Blockly.getMainWorkspace();
 
             let xml = Blockly.Xml.workspaceToDom(workspace);
             let xml_text = Blockly.Xml.domToText(xml);
 
-            window.top.postMessage({ type: "loadSample", code: xml_text }, "*");
+            window.parent.postMessage(
+              { type: "loadSample", code: xml_text },
+              "*"
+            );
           }
         },
         false
--- a/web/blockframe.html
+++ b/web/blockframe.html
@@ -47,7 +47,6 @@
 
     <script src="../js/blocks_runmain.js"></script>
     <script>
-      let upperParent = null;
       let frameId = null;
 
       function loadXml(data) {
@@ -70,10 +69,6 @@
 
       function setId(j_id_param) {
         frameId = j_id_param;
-      }
-
-      function setParent(par) {
-        upperParent = par;
       }
 
       document.getElementById("blocklyDiv").addEventListener(
--- a/web/blocks.html
+++ b/web/blocks.html
@@ -49,7 +49,7 @@
             <a id="saveAsButton" class="cw-button blue">
               <i class="mdi mdi-18px mdi-dots-horizontal"></i>&nbsp; Save As
             </a>
-            <a id="deleteButton" class="cw-button red" style="display: none;">
+            <a id="deleteButton" class="cw-button red">
               <i class="mdi mdi-18px mdi-delete"></i>&nbsp; Delete
             </a>
           </div>
@@ -110,7 +110,7 @@
           <a id="signin" class="cw-button blue"
             ><i class="mdi mdi-18px mdi-login"></i>&nbsp; Sign In</a
           >
-          <a id="signout" class="cw-button blue" style="display: none;"
+          <a id="signout" class="cw-button blue"
             ><i class="mdi mdi-18px mdi-logout"></i>&nbsp; Sign Out</a
           >
           <a id="docButton" class="cw-button blue"
--- a/web/blocks.html
+++ b/web/blocks.html
@@ -37,33 +37,19 @@
       <div class="ui-layout-container">
         <div id="nav" class="dropbox ui-layout-west">
           <div class="nav_items">
-            <a id="newFolderButton" class="cw-button red" onclick="newFolder()">
+            <a id="newFolderButton" class="cw-button red">
               <i class="mdi mdi-18px mdi-folder"></i>&nbsp; New Folder
             </a>
-            <a id="newButton" class="cw-button red" onclick="newProject()">
+            <a id="newButton" class="cw-button red">
               <i class="mdi mdi-18px mdi-plus"></i>&nbsp; New
             </a>
-            <a
-              id="saveButton"
-              class="cw-button blue"
-              style="display: none;"
-              onclick="saveProject()"
-            >
+            <a id="saveButton" class="cw-button blue" style="display: none;">
               <i class="mdi mdi-18px mdi-cloud-upload"></i>&nbsp; Save
             </a>
-            <a
-              id="saveAsButton"
-              class="cw-button blue"
-              onclick="saveProjectAs()"
-            >
+            <a id="saveAsButton" class="cw-button blue">
               <i class="mdi mdi-18px mdi-dots-horizontal"></i>&nbsp; Save As
             </a>
-            <a
-              id="deleteButton"
-              class="cw-button red"
-              style="display: none;"
-              onclick="deleteProject()"
-            >
+            <a id="deleteButton" class="cw-button red" style="display: none;">
               <i class="mdi mdi-18px mdi-delete"></i>&nbsp; Delete
             </a>
           </div>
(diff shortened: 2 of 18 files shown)


web/js/codeworld_shared.js
```diff
@@ -1541,11 +1367,17 @@ function preFormatMessage(msg) {
   msg = Html.encode(msg)
     .replace(
       /program\.hs:(\d+):((\d+)(-\d+)?)/g,
-      '<a href="#" onclick="goto($1, $3); return false;">Line $1, Column $2</a>'
+      `<a href="#" onclick=${() => {
```
cdsmith, line 1370: Can you explain how this works?  I'm confused.  We're building a string here, not a DOM element, so how can you embed a literal function into it?

web/js/codeworld_shared.js
```diff
@@ -2004,15 +1836,337 @@ function updateChildrenIndexes(node) {
...
+  if (window.savedGeneration === null) return doc.getValue() === '';
+  else return doc.isClean(window.savedGeneration);
+}
+
+/*
+ * Updates all UI components to reflect the current state.  The general pattern
+ * is to modify the state stored in variables and such, and then call updateUI
+ * to get the visual presentation to match.
+ */
+function updateUI() {
```
cdsmith, line 1874: This definitely feels like the wrong place for this function.  The point of this function is to adjust visibility of specific elements on the page.  `codeworld_shared.js` is the code that's shared by multiple different pages (blocks.html and env.html), which have different elements on them.  And it looks like blocks still has its own `updateUI` as well?

What about this?  Instead of having one updateUI function, have codeworld_shared.js export a function to add an event handler, and fire events instead of calling updateUI directly.  Then each environment can provide its own updateUI function that does the right things for that page.

web/js/funblocks.js
```diff
@@ -290,10 +315,9 @@ function compile(src, silent) {
  * to get the visual presentation to match.
  */
 function updateUI() {
```
cdsmith, line 317: So if I understand correctly, the updateUI in codeworld_shared is still used only by codeworld.js, and not by funblocks.js?  Or will calls to updateUI from within codeworld_shared.js still call the local one, but calls from here will call the override here?  That's quite confusing...

web/js/codeworld.js
```diff
@@ -762,24 +727,6 @@ function help() {
...
-    return true;
-  }
-
-  const doc = window.codeworldEditor.getDoc();
-
-  if (window.savedGeneration === null) return doc.getValue() === '';
-  else return doc.isClean(window.savedGeneration);
-}
-
-function loadSample(code) {
```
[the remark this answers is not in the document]
Powell-v2: Thanks for pointing that out, I overlooked it.

`window.postMessage` doesn't allow sending functions, unless you stringify them and use `eval` on the other side. Not sure this is desirable?

web/js/codeworld_shared.js
Powell-v2: I got mixed up, interpolating a function here wouldn't work. Will revert this change.

web/js/codeworld.js
```diff
@@ -35,7 +105,12 @@ async function init() {
   preloadBaseBundle();
   window.setInterval(preloadBaseBundle, 1000 * 60 * 60);
 
-  initDirectoryTree();
+  function loadProject(name, path) {
```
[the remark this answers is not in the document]
Powell-v2: Good point, I'll take care of it.

web/js/codeworld.js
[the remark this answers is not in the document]
cdsmith: Why can't you just replace `parent.loadSample(theCode)` with `parent.postMessage({ type: 'loadSample', code: theCode }, '*')`?  Then instead of defining `loadSample` in funblocks.js and codeworld.js, you'd need to add a case to [the message event handler](https://github.com/google/codeworld/blob/30ff2f7a0a0564ef295121d593362331d863c257/web/js/codeworld.js#L888)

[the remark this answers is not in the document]
Powell-v2: I haven't noticed this piece of code before. Thanks for referencing it, that would definitely work!

web/blockframe.html
```diff
@@ -79,13 +79,13 @@
       document.getElementById("blocklyDiv").addEventListener(
         "click",
         () => {
-          if (upperParent) {
+          if (window.top) {
```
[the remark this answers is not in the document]
Powell-v2: Yes, should work, will correct it.

web/env.html
```diff
@@ -158,38 +132,22 @@
...
-            onclick="inspect()"
+          <a id="inspectButton" class="cw-button cyan"
             ><i class="mdi mdi-18px mdi-magnify"></i>&nbsp; Inspect</a
           >
-          <a class="cw-button red" onclick="stopRun()"
+          <a id="stopButton" class="cw-button red"
             ><i class="mdi mdi-18px mdi-stop"></i>&nbsp; Stop</a
           >
-          <a class="cw-button green" onclick="compile()"
+          <a id="startButton" class="cw-button green"
```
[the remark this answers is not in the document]
Powell-v2: Sorry, you mentioned that already.

web/js/codeworld.js
```diff
@@ -13,17 +13,144 @@
...
+    if (window.debugAvailable) {
+      $inspectButton.show();
+    } else {
+      $inspectButton.hide();
+    }
+  });
+
+  onObjectPropertyChange(window, 'debugActive', () => {
+    if (window.debugActive) {
+      $inspectButton.css('color', 'black');
```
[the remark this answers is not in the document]
Powell-v2: Yes, toggling a class is a more preferable choice, will refactor. I just didn't want to introduce any unrelated changes.

web/css/codeworld.css
```diff
@@ -116,6 +116,26 @@ body {
   margin: 0px;
 }
 
+#navButton {
```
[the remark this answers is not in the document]
Powell-v2: Why does it seem weird? Any reason for keeping it inline?

web/js/utils/object.js
```diff
@@ -0,0 +1,15 @@
+const onObjectPropertyChange = (object, property, callback) => {
```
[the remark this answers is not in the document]
Powell-v2: Yes, indeed, the current implementation allows defining only one callback at once. I'll expand it to handle any amount of listeners, following your advice.

web/blocks.html
```diff
@@ -37,33 +37,19 @@
...
+            <a id="newButton" class="cw-button red">
               <i class="mdi mdi-18px mdi-plus"></i>&nbsp; New
             </a>
-            <a
-              id="saveButton"
-              class="cw-button blue"
-              style="display: none;"
-              onclick="saveProject()"
-            >
+            <a id="saveButton" class="cw-button blue">
```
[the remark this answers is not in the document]
Powell-v2: Yes, for the sake of consistency, delete button's style should also be transferred to CSS. I simply overlooked it, thanks for pointing it out!

web/blockframe.html
```diff
@@ -79,13 +79,16 @@
       document.getElementById("blocklyDiv").addEventListener(
         "click",
         () => {
-          if (upperParent) {
+          if (window.parent) {
```
cdsmith, line 82: It looks like `upperParent` is no longer used.  Let's remove it.

web/css/codeworld.css
```diff
@@ -116,6 +116,20 @@ body {
   margin: 0px;
 }
 
+#signout,
```
cdsmith, line 119: I'm still not sold on these being part of the stylesheet, to be honest.  They don't feel like they are part of the layout, so much as part of the logic.  They will be modified in JavaScript, which means there's something weird going on, where the initial state (no element-level CSS, and `display` inherited from the stylesheet) is different from the future states (`display: none` set on the element)

Not a big deal, but I found it clearer when elements were always shown or hidden using element-level styles.

web/env.html
```diff
@@ -45,40 +45,22 @@
...
+            <a id="newButton" class="cw-button red">
               <i class="mdi mdi-18px mdi-plus"></i>&nbsp; New
             </a>
-            <a
-              id="saveButton"
-              class="cw-button blue"
-              style="display: none;"
-              onclick="saveProject()"
-            >
+            <a id="saveButton" class="cw-button blue">
```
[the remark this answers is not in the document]
cdsmith: Discussed elsewhere.  I'll leave that conversation open, and whatever is decided there should be applied consistently.

web/js/codeworld.js
```diff
@@ -13,17 +13,148 @@
...
+  $('#inspectButton').on('click', inspect);
+
+  $('#runButton').on('click', compile);
+  $('#stopButton').on('click', stopRun);
+
+  $('#runner').on('load', toggleObsoleteCodeAlert);
+}
+
+function attachCustomEventListeners() {
+  $('#nav').on('enable', () => {
```
cdsmith, line 90: I think this is what I've been trying to avoid, actually.  I think it's more important to get this merged than to get everything right, so I'm not insisting on changes right now.  But I think, like in the past, I really want one source of truth for everything, and the source of truth for whether you're signed in should be `Auth.signedIn()`.

So instead of saying "when nav is enabled, do these things", we should really say "when you're signed in, do these things".  The logic should be idempotent, so calling it twice is fine.  Then you could add a listener that gets called whenever the signed-in state changes, that runs that code.

The `onObjectPropertyChange` sections below are perfect, and get this right.  The underlying state changes should drive UI changes.  UI changes shouldn't trigger other UI changes.

But, not a big deal to do all this right away.

web/js/codeworld.js
```diff
@@ -1261,83 +1251,6 @@ function downloadProject() {
...
-  });
-  return errors;
-}
-
-function clearWorkspace() {
-  utils.directoryTree.clearSelectedNode();
-
-  setCode('');
-}
+// TEMP: required by setCode in codeworld_shared.js
```
cdsmith, line 1254: Do you want to merge first and then fix this?  Or is this a work in progress?

web/js/utils/auth.js
```diff
@@ -1,21 +1,7 @@
-/*
```
cdsmith, line 1: We do need to keep these headers?  Google requires them on code hosted under its GitHub account.  Make sure they exist on any new source files that you add, as well.

web/css/codeworld.css
Powell-v2: I see your point, it makes sense. I could restore inline styles - let me know if that's how you want me to proceed.

cdsmith: Yeah, let's move just the `display` styles back to inline, please.

web/js/codeworld.js
Powell-v2: I probably misunderstood your intention to implement custom events. `$('#nav').trigger('event_name')` simply dispatches an event, there are implied UI actions when calling this method.

Actually, we already define a listener that gets called whenever user's logged in state changes - I moved respective actions there.

web/js/codeworld.js
Powell-v2: Let's prioritize merging this PR. These are things that I can take care of later (blocks page has few of those globally defined methods as well).

PR #1392, opened 2020-07-06
Append output instead of entirely replacing it every time.

### LINKED ISSUE #1389: Long console output gets very slow
**Shared Program**: https://code.world/haskell#PpcrjDqEzZaVUG0a4_0b_Kg

**Description**

If a console program prints a lot of output, it gets very slow.  The reason seems to be that instead of just appending the content to the existing output element, the entire last output element is rebuilt from scratch.  This looks to be part of the collapsible UI code that was built for error reporting, but inexplicably includes output as well.

We should special-case normal output from programs, so that it is appended directly without rebuilding the element.  If we could remove the collapsible UI at the same time, that's even better!

Powell-v2:
Closes #1389

### DIFF
--- a/web/js/codeworld_shared.js
+++ b/web/js/codeworld_shared.js
@@ -1560,56 +1560,62 @@ function preFormatMessage(msg) {
 
 function printMessage(type, message) {
   const $outputBlock = $('#message');
+  const $lastOutputBlock = $outputBlock.children().last();
 
-  let $box = $outputBlock.find('.message-box');
-  let $messageContent = $outputBlock.find('.message-content');
+  const formattedMessage = preFormatMessage(message);
+  const lines = formattedMessage.trim().split('\n');
 
-  if (!$box.length && !$messageContent.length) {
-    $box = $('<div>');
+  // Combine sequential log messages.
+  if (
+    type === 'log' &&
+    $lastOutputBlock.length &&
+    $lastOutputBlock.attr('class').includes('log')
+  ) {
+    const $messageContent = $lastOutputBlock.find('.message-content');
+
+    if ($messageContent.children().length) {
+      $messageContent.append(lines);
+    } else {
+      $messageContent.replaceWith(
+        `${'<details open="open" class="message-content">' +
+          `<summary>${$messageContent.html()}</summary>`}${ 
+          lines 
+        }</details>`
+      );
+    }
+  } else {
+    const $box = $('<div>');
     $box.addClass(`message-box ${type}`);
 
     const $messageGutter = $('<div>');
     $messageGutter.addClass('message-gutter');
 
-    $messageContent = $('<div>');
-    $messageContent.addClass('message-content');
+    const $messageContent = $('<div>');
+    $messageContent.addClass('message-wrapper');
 
     $box.append($messageGutter, $messageContent);
     $outputBlock.append($box);
-  }
 
-  const formattedMessage = preFormatMessage(message);
-  const lines = formattedMessage.trim().split('\n');
+    if (lines.length < 2) {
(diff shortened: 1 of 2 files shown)


web/js/codeworld_shared.js
```diff
@@ -1559,78 +1559,84 @@ function preFormatMessage(msg) {
 }
 
 function printMessage(type, message) {
-  const outputDiv = document.getElementById('message');
+  const $outputBlock = $('#message');
```
cdsmith, line 1562: Why all the '$' prefixes for variables?

web/js/codeworld_shared.js
```diff
@@ -1559,78 +1559,84 @@ function preFormatMessage(msg) {
...
-  } else {
-    box = document.createElement('div');
-    box.classList.add('message-box');
-    box.classList.add(type);
+  let $box = $outputBlock.find('.message-box');
+  let $messageContent = $outputBlock.find('.message-content');
 
-    const messageGutter = document.createElement('div');
-    messageGutter.classList.add('message-gutter');
+  if (!$box.length && !$messageContent.length) {
```
cdsmith, line 1567: I think the logic is wrong here.  If the type is 'log', and the last output type is also 'log', then this should be appended.  Otherwise, it should be added as a new box.

In particular:
1. If a 'log' message comes after an 'error' or 'warning' message, it should be in a new box.
2. If there are two 'error's or 'warning's in a row, they should still be in separate boxes.
3. ONLY if there's two 'log's in a row should they be combined.

web/js/codeworld_shared.js
```diff
@@ -1559,78 +1559,84 @@ function preFormatMessage(msg) {
...
+  const formattedMessage = preFormatMessage(message);
+  const lines = formattedMessage.trim().split('\n');
 
-  messageContent.innerHTML = '';
+  let $details = $messageContent.find('details');
+  let $summary = $messageContent.find('summary');
+  let $firstLine;
 
-  let firstLine;
   if (lines.length < 2) {
```
cdsmith, line 1588: I'm not sure this will correctly handle the transition from a single-line message to a multi-line message, when new content is appended.

web/js/codeworld_shared.js
Powell-v2: This prefix indicates that given DOM element was spawned (or selected) by jQuery and has all associated methods attached to it. It's fairly common to do so, but if you don't like it, I'll remove it.

web/js/codeworld_shared.js
Powell-v2: I haven't considered this at all. Will correct the logic, thanks for clarifying!

web/js/codeworld_shared.js
```diff
@@ -1559,78 +1559,101 @@ function preFormatMessage(msg) {
...
+    $lastOutputBlock.length &&
+    $lastOutputBlock.attr('class').includes('log')
+  ) {
+    const $messageContent = $lastOutputBlock.find('.message-content');
 
-    messageContent = document.createElement('div');
-    messageContent.classList.add('message-content');
+    if ($messageContent.children().length) {
+      $messageContent.append(lines);
+    } else {
```
cdsmith, line 1578: Will this branch ever get taken?  You're always adding at least one child to any element.

Powell-v2: Yes, it will be in case of a single-line message as it contains only text:

```html
<div>
  {lines}
</div>
```

On the other hand, multiline messages always contain a `summary` tag, and that's where we'd like to inject the report link:

```html
<details>
  <summary>
    {first_line}
  </summary>
  {remaining_lines}
</details>
```

PR #1400, opened 2020-07-24
WIP: Redirects for autocomplete

### LINKED ISSUE #1380: Redirects for autocomplete
Right now, the autocomplete list is populated based on a weighted edit distance between what's been typed already and the symbol being considered.  This works well when a student knows part or all of the function name they are looking for, but it misses cases where a student wants to draw a square, or a triangle.  There is no `square` function, but squares can be drawn using `rectangle`.  There is no `triangle`, but triangles can be drawn using `polygon`.

I propose to add this knowledge in the form of redirects for autocomplete, so that if a student triggers autocomplete with something close to `square`, we should suggest `rectangle` as a completion.  Similarly, `triangle` should redirect to `polygon`.  Maybe `ellipse` and `oval` should redirect to `scaled(circle`?  Though it's weird to add unbalanced parentheses.

Powell-v2:
Closes #1380


web/js/codeworld_shared.js
```diff
@@ -532,13 +532,36 @@ function registerStandardHints(successFunc) {
...
     }
 
-    const options = found[0].concat(found[1]).concat(found[2]);
+    const mappedTerms = {
+      square: 'rectangle',
+      triangle: 'polygon',
+    };
+
+    for (const [lookedUpTerm, substitute] of Object.entries(mappedTerms)) {
+      if (lookedUpTerm.startsWith(term)) {
```
cdsmith, line 541: You seem to be disabling fuzzy matching for substitutions here.  Any reason?  I can definitely see students misspelling square or triangle.  I think if you just leave out the `if` statement, the fuzzy scoring will already do the right thing.  It already scores prefix matches much higher than others.

web/js/codeworld_shared.js
```diff
@@ -532,13 +532,36 @@ function registerStandardHints(successFunc) {
...
-      candidate.cost = substitutionCost(
-        token.string,
-        candidate.text,
-        term.length
-      );
+      const { isSubstitute, originalTerm, text } = candidate;
+
+      if (isSubstitute) {
+        candidate.cost =
+          substitutionCost(token.string, originalTerm, term.length) + 5;
```
cdsmith, line 561: Just a minor style nit, but I'd rather this penalty be added inside the scoring function, so that all of the cost logic is in one place.

web/js/codeworld_shared.js
```diff
@@ -532,13 +532,36 @@ function registerStandardHints(successFunc) {
       found[2] = [];
     }
 
-    const options = found[0].concat(found[1]).concat(found[2]);
+    const mappedTerms = {
+      square: 'rectangle',
```
cdsmith, line 536: It would be slightly more correct to map these to `Prelude.rectangle` and `Prelude.polygon`.  That's still not 100% right, since someone could presumably do:

    import Prelude ()
    import qualified SomeOtherModule as Prelude

but we can assume that anyone doing that is trying to fool the system, so that's fine.  But I can definitely see someone in, for example, LSU's class that uses their own standard library running into problems with this mapping.

web/js/codeworld_shared.js
```diff
@@ -532,13 +532,36 @@ function registerStandardHints(successFunc) {
...
+      triangle: 'polygon',
+    };
+
+    for (const [lookedUpTerm, substitute] of Object.entries(mappedTerms)) {
+      if (lookedUpTerm.startsWith(term)) {
+        found[2].push({
+          text: window.codeWorldSymbols[substitute].insertText,
```
cdsmith, line 543: You should check that window.codeWorldSymbols[substitute] exists.  I can arrange for it not to (for instance, but adding `import Prelude()` to the top of the file), and I believe your code will throw exceptions in that case.

web/js/codeworld_shared.js
Powell-v2: No particular reason. I didn't realize that I was disabling fuzzy matching logic, and we should certainly keep it. Thanks for pointing that out!

web/js/codeworld_shared.js
```diff
@@ -402,12 +402,13 @@ function getQualifierPrefix(cm, pos) {
...
 }
 
-function substitutionCost(a, b, fixedLen) {
+function substitutionCost(a, b, fixedLen, isTermReplaced) {
   const insertCost = 1;
   const deleteCost = 1.5;
   const transCost = 1;
   const substCost = 1.5;
   const caseCost = 0.1;
+  const penalty = 5;
```
cdsmith, line 411: Let's call this `redirectPenalty` so it's clearer what it means without reading the code later.

web/js/codeworld_shared.js
```diff
@@ -532,12 +537,48 @@ function registerStandardHints(successFunc) {
...
+        matchingCandidate = bucket.find(({ text }) =>
+          replacementTerm.includes(text)
+        );
+
+        if (matchingCandidate) {
+          break;
+        }
+      }
+
+      if (window.codeWorldSymbols[replacementTerm] && !matchingCandidate) {
```
cdsmith: You can check `window.codeWorldSymbols[replacementTerm]` and `continue;` at the beginning of the loop.  It will save work for cases where that's false.

cdsmith: I'm confused about why you're checking `!matchingCandidate`.  In fact, I've pretty much convinced myself that `matchingCandidate` will always be true if the symbol is in `codeWorldSymbols`.  We just finished iterating over every key in `codeWorldSymbols` and adding it to `found`.  Am I wrong?

That's actually an interesting point, though.  All of these completions **already** exist in `found[2]`, so maybe the task here isn't to add a new completion, but to adjust the scores to promote them.  That is, the ultimate score for a choice in bucket 2 should be the maximum score of any target term that redirects to this one (after applying the redirect penalty).  That's a completely different implementation, so... sorry about that.  Worth thinking about, though.

web/replacement_terms.json
```diff
@@ -0,0 +1,71 @@
+{
```
cdsmith, line 1: Thinking about the scoring step, I think you actually want to invert this file, so it's something like this:

```
{
  "abs": ["positive", "magnitude"],
  "acos": ["arccos"],
  ...
  "polygon": [
    {
      "value": "triangle",
      "explanation": "A triangle is a polygon with three vertices."
    },
    "quadrilateral",
    "parallelogram",
    "trapezoid",
    "rhombus",
    "pentagon",
    "hexagon",
    "heptagon",
    "octagon",
  ],
  ...
}
```

cdsmith:
We lost the protection again matching the wrong symbols here.  I've just pushed a simple (one-line) change to add a `definingModule` property to `codeWorldSymbols` entries.  We should ensure that the redirects here are only applied to symbols in the right module.  (Right now., that module is always `Prelude`.  However, one should be able to add redirects to the file for symbols in non-Prelude modules.  I already anticipate adding some for the `Sketches` module.)

That will avoid this weird corner case:

```
import Prelude hiding (rectangle)
rectangle = polygon([(1, 1), (1, 3), (2, 3), (2, 1)])
```

With the current implementation, `square` would still autocomplete to `rectangle`, even though the `rectangle` in scope here isn't  the one from the standard library.  You can now detect this, because in this case, `window.codeWorldSymbols['rectangle']` will be missing its `definingModule` property, so you'll know it's not the right one.

web/js/codeworld_shared.js
```diff
@@ -449,12 +457,22 @@ function substitutionCost(a, b, fixedLen) {
...
 }
 
 // Hints and hover tooltips
 function registerStandardHints(successFunc) {
-  CodeMirror.registerHelper('hint', 'codeworld', (cm) => {
+  let replacementTerms;
+
+  CodeMirror.registerHelper('hint', 'codeworld', async (cm) => {
+    if (!replacementTerms) {
+      const blob = await fetch('./replacement_terms.json');
```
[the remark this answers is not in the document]
Powell-v2: Yes, there is a chance on a slow connection that there will be a perceptible delay when invoking the autocompletion for the first time. The thing with `registerStandardHints` is that it's getting called by the guide frame, and the list with replacement terms would be downloaded again when the guide is requested.

I'm not sure how to update the autocompletion list once it's shown to the user. Do I have to manually call the hint helper?

web/js/codeworld_shared.js
```diff
@@ -532,12 +550,42 @@ function registerStandardHints(successFunc) {
...
-    const options = found[0].concat(found[1]).concat(found[2]);
+    for (const [replacementTerm, lookedUpTerms] of Object.entries(
+      replacementTerms
+    )) {
+      lookedUpTerms.forEach((lookedUpTerm) => {
+        const withOptions = typeof lookedUpTerm === 'object';
+
+        if (
+          window.codeWorldSymbols[replacementTerm] &&
+          window.codeWorldSymbols[replacementTerm].definingModule
```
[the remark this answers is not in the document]
Powell-v2: Got it. I'll include the module name, no problems.

web/js/codeworld_shared.js
```diff
@@ -532,12 +550,42 @@ function registerStandardHints(successFunc) {
...
+        ) {
+          found[2].push({
```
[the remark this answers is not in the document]
Powell-v2: Yes, I'm working on it now.

web/js/codeworld_shared.js
```diff
@@ -449,11 +457,18 @@ function substitutionCost(a, b, fixedLen) {
...
+    d[b.length][a.length] +
+    scale(fixedLen) * (a.length - b.length) +
+    (isTermReplaced ? redirectPenalty : 0)
+  );
 }
 
 // Hints and hover tooltips
-function registerStandardHints(successFunc) {
+async function registerStandardHints(successFunc) {
+  const blob = await fetch('./replacement_terms.json');
```
cdsmith, line 469: Why not do something like this:

```
let replacementTerms = { definingModule: "", mapping: {} };
fetch('./replacement_terms.json').then((blob) => {
  replacementTerms = await blob.json();
});
```

May as well avoid the extra startup delay.

web/js/codeworld_shared.js
```diff
@@ -482,46 +497,52 @@ function registerStandardHints(successFunc) {
...
           };
-          if (hint === prefix + token.string) {
+          if (hintName === prefix + token.string) {
             found[0].push(candidate);
-          } else if (hint.startsWith(prefix + term)) {
+          } else if (hintName.startsWith(prefix + term)) {
             found[1].push(candidate);
           } else {
-            found[2].push(candidate);
+            if (hintProps.definingModule === replacementTerms.definingModule) {
```
cdsmith, line 525: I don't think you want this change.  In this branch, the hint is for an entire module.  Replacement terms are irrelevant in that case.  (There are no replacement terms for modules, only for identifiers.)  Module candidates also won't have a `definingModule` property, since it doesn't make sense.

web/js/codeworld_shared.js
```diff
@@ -482,46 +497,52 @@ function registerStandardHints(successFunc) {
...
+            renderDeclaration(elem, hintProps, 50);
           },
         };
         if (hintIdent === token.string) {
           found[0].push(candidate);
         } else if (hintIdent.startsWith(term)) {
           found[1].push(candidate);
         } else {
-          found[2].push(candidate);
+          if (hintProps.definingModule === replacementTerms.definingModule) {
```
cdsmith, line 543: I don't think this is right, either.  You still want to add the candidate under its original name, even if it isn't in the right module for replacement terms to apply.

web/js/codeworld_shared.js
```diff
@@ -532,14 +553,49 @@ function registerStandardHints(successFunc) {
...
+      const originalTermCost = substitutionCost(
         token.string,
-        candidate.text,
+        text,
         term.length
       );
-    }
+
+      const mappedTerms =
+        Object.prototype.hasOwnProperty.call(replacementTerms.mapping, text) &&
```
cdsmith, line 567: This looks strangely complex.  Why not just `replacementTerms.mapping.hasOwnProperty(text)`?

web/js/codeworld_shared.js
```diff
@@ -532,14 +553,49 @@ function registerStandardHints(successFunc) {
...
+      const { text } = candidate;
+      const originalTermCost = substitutionCost(
         token.string,
-        candidate.text,
+        text,
         term.length
       );
-    }
+
+      const mappedTerms =
```
cdsmith, line 566: This is where the filtering by defining module should go, I think.  If the candidate has no defining module, or if its defining module doesn't match `replacementTerms`, then you should skip this rescoring logic.

web/js/codeworld_shared.js
```diff
@@ -554,11 +610,18 @@ function registerStandardHints(successFunc) {
           break;
         }
       }
-      const goodOptions = options.slice(0, numGood);
+
+      const goodOptions = options
+        .slice(0, numGood)
+        .reduce((result, current) => {
```
cdsmith, line 616: What's the problem you are trying to solve here?

web/replacement_terms.json
```diff
@@ -0,0 +1,70 @@
...
+  "definingModule": "Prelude",
```
cdsmith, line 2: What about:

```
{
  "Prelude": {
    "abs": ["positive", "magnitude"],
    ...
    "translated": ["move"]
  },
  "SomeOtherFutureModule": {
    ...
  }
}
```

I don't think this is really any more work than the single-module implementation that you have now.  For now, you only have replacements for `Prelude`, but I can add more modules later.

web/js/codeworld_shared.js
Powell-v2: Okay, I'll remove this condition. Thanks for the explanation.

web/js/codeworld_shared.js
Powell-v2: Okay, I'll remove this condition. Thanks for the explanation.

web/js/codeworld_shared.js
Powell-v2: Yeah, looks odd, but I had to do it since ESLint complained ([link to the respective rule](https://eslint.org/docs/rules/no-prototype-builtins)).

cdsmith: I see.  Okay, that looks like there's some justification, so let's leave it that way.

web/js/codeworld_shared.js
Powell-v2: This servers no purpose any longer, will remove it.

web/replacement_terms.json
Powell-v2: Sure, no problems, everything could be kept in one file.

Powell-v2:
Is there anything left to be done here?

web/js/codeworld_shared.js
Powell-v2: What do you have in mind as a replacement?

Powell-v2: Isn't `async/await` just a syntactic sugar over promises?

cdsmith: My understanding is that if you `await` at the beginning of `registerStandardHints`, then the rest of this function will not run until the fetch is complete.  That means we won't have any autocomplete at all, won't even start loading `codeworld-base.txt`, etc., until this fetch is complete.  That's not really needed.  It's okay to run the rest of this function without waiting on that network round trip.

Powell-v2: Yes, you are right, code execution literally pauses at the line containing the `await` keyword and continues only once the async operations is complete. There is no such issue with promises. Turns out I had a misconception about the way `await` works ... Now it's cleared up. Will update the code.

PR #1403, opened 2020-08-01
Reorganize guide's navbar.

### LINKED ISSUE #1022: Group Extras javadoc together in Guide
The documentation for various Extras modules is taking up a bit too much space in the Guide.  We want it there, but in a single item until expanded.

Powell-v2:
Closes #1022

### DIFF
--- a/web/css/doc.css
+++ b/web/css/doc.css
@@ -103,32 +103,40 @@ pre {
   background: var(--doc-tr-odd-bg);
 }
 
-#helpacc {
-  position: fixed;
-  overflow: auto;
-  width: 19em;
+.accordion {
   height: 100%;
-  margin: 0px 0px 0px 0px;
-  padding: 0px 0px 0px 0px;
-  border-right: solid var(--doc-helpacc-border-right) 3px;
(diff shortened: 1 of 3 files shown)


Powell-v2:
@cdsmith this one is ready for you to review.

cdsmith:
@Powell-v2 

This code looks good, and I merged it.  However, while testing, I realized that saving the current help document position has been broken for a while.  Specifically, when the help window is closed, the parent does an `iframe.contentWindow.savePosition()`, has been failing because `doc.js` no longer adds `savePosition` function to the `window` object.  I've fixed that with a quick hack, but if there's a better answer, happy to do it that way.

Powell-v2:
I think it's alright to keep the function on the `window` object. Another option would be to communicate via `postMessage()`.

cdsmith:
Let's leave it as-is.  I thought about `postMessage`, but my concern is that since the window is being closed, an async message may not be received before the document is destroyed.  That could be worked around, but it seems tricky.  Let's leave it as is.

Issue #1407, opened 2020-08-05
Local instance loads remote Web resources


jbash:
I can't call this a bug report, because it's obviously intentional, but the CodeWorld home page of a locally installed instance loads may remote resources, including--

+ Google APIs (with Google authentication disabled)
+ JQuery from jquery.com
+ Something from jsdelivr.net
+ Something from wurfl.io
+ Something from polyfill.io
+ (Presumably) icons from materialdesignicons.com

I would not expect a local instance to load any code from, or communicate in any way with, anything but its own server, and definitely not without explicit permission from an administrator.

cdsmith:
This is something I'd definitely love to fix if someone wants to work on it.  There are two different parts:

For jquery.com, jsdelivr.net, materialdesignicons.com, and ajax.googleapis.com, the right solution is to copy the relevant version into third_party, along with any license files, then symlink from web/js to the script in third_party.  (#1374 may change some details of this, but that change is not imminent.)  Note that the googleapis.com isn't actually related to Google-based login; it's just the canonical location of a CSS file that's part of jquery-ui.

Something different will need to be done for polyfill.io and wurfl.io.  These are services that serve different content based on the HTTP client headers, so we're actually relying on server-side logic from these web services.  Specifically: wurfl.io is used to detect mobile browsers in order to work around a bug in CodeMirror, and polyfill.io is used to backfill some libraries that are used by the CodeWorld code and aren't implemented in older browsers.  If there's a solution to these problems that doesn't use a web service, great.  I don't know so much about the JavaScript ecosystem.  @Powell-v2 is helping out and is better informed than I am, so perhaps he has some ideas.

Powell-v2:
`WURFL` could be replaced with a simplified check on the client. One plausible solution is to detect whether a device is touchscreen while looking at the screen width at the same time. This should sift out most of the phones / tablets. The question here is: how accurate this detection has to be? What this bug in Codemirror is about, actually? Maybe it's been resolved already? I can see that the lib was added back in 2017.

`polyfill.io` could be replaced with a static file. Or, perhaps, we can remove it altogether? Browsers have caught up during past several years and most of the required features are supported natively. There is one notable exception: IE, but Code.World doesn't work there these days anyway due to the ES6 syntax.

cdsmith:
[...]

PR #1412, opened 2020-08-10
Serve static third-party code from local codeworld server (download at build time)

jbash:
This is a partial fix for #1407; it deals with most references to purely static remote code like JQuery and icons, so that users' browsers don't end up downloading those resources from servers other than the one they're actually visiting.

 It does not eliminate any of the following more difficult cases:

1. Script embeds from polyfill.and or wurfl.io (and polyfill.io's "internal" reference to fastly.net). These rely on server-side logic to dynamically generate the loaded code in browser-dependent ways, so they can't easily be moved to a local CodeWorld server.

2. Some CodeWorld programs embedded in iframes in documentation. These use explicit references to the main server at code.world. Although these aren't "foreign" embeds on code.world, they are "foreign" on private instances.

3. Some GitHub feedback links. In addition to being ugly, these apparently only work properly on the main code.world server. However, they're not fetched unless they're actually clicked.

It's inconvenient to actually check this third-party code into the CodeWorld repository; Google policy on redistributing code leads to unnecessary clerical work, and updates to new versions of the third party code require unnecessary local Git fiddling. Instead, this patch takes the approach of downloading the third-party code at CodeWorld install time (from install.sh). SHA-256 hashes make sure that we're getting the exact third-party code that we expect.

Even though these changes don't actually add anything to the third_party subtree, they still clean up the Git pre-commit hook to eliminate linting and pretty printing on any code under that subtree.

On the other hand, these changes do *not* deal with another conceptually separate problem found during investigation of #1407: blocks.html still includes two separate versions of JQuery, but probably only uses the first. It seemed safer to save that issue for a separate change.

This stuff was tested interactively by using the server and tracing network downloads, but does not include any automated test cases.


jbash:
@googlebot I signed it!

cdsmith:
(Sorry I didn't get a chance to look at this last night.  I'll take a look tonight.)

install.sh
```diff
@@ -173,6 +173,15 @@ if [ ! -f $BUILD/progress/system-pkgs ]; then
   touch $BUILD/progress/system-pkgs
 fi
 
+# Locally mirror various third-party JavaScript, CSS, etc, so that we don't
```
cdsmith, line 176: If you're going to do it this way, can we do it at build time?  This is okay if it takes a long time, but rerunning install.sh isn't very common, and if it's quick., I'd rather just do it at build.

mirrored/.gitignore
```diff
@@ -0,0 +1 @@
+*
```
cdsmith, line 1: Doesn't this exclude your own scripts from source control?  I'm confused.

mirrored/get_mirrored
```diff
@@ -0,0 +1,37 @@
...
+    )
+MASTER_HASHES=sums.txt
+
+shopt -s extglob
+
+exitstat=0
+
+for url in "${TO_MIRROR[@]}"; do
+    file=".${url#http?(s):/}"
+    [ -f "${file}" ] || wget --no-verbose --force-directories "${url}" || exitstat=1
```
cdsmith, line 25: Anything fetched or built during install/build should go in the `build` directory.

jbash:
> In general, I've been pondering this, and I'm still not sure of this approach. If there are technical reasons for it, then can we talk about what they are? [...] In general, it's better to comply with organizational guidelines than try to work around them by doing something more complex.

I don't see how this is more complex than the alternative. In fact, I think it's much simpler, and the main reasons for doing it are reasons of simplicity.  I don't know if they count as "technical reasons" or not.

The reasons:

1. It gives you an intuitve code tree that most experienced people will understand instantly.

    Since I'm not committing to maintain this, I wanted to make sure it was easy for the next person, whether that was you or anybody else, to update versions or add mirrored material as necessary.

    With this structure, you can mirror something with a one line change in a single file.

     In a more complicated structure with arbitrary project names and license files, not only do you have to go find out where the file you want to mirror actually comes from, but you have to come up with your own idea for where to put it locally. Then you also have to chase down the license for it and figure out exactly where you're required to put that, which is intrinsically manual and can take some time, especially if you're new at it.

    Anything that discourages taking updates, or that discourages new contributors, seems like a bad idea to me.

2. It reduces the number of objects you have to track, and it works without either having "multiple sources of truth" or having to master relatively arcane things like Git submodules.

    If third-party code is checked in locally, and you find a file there, you have to figure out whether it's been modified within the CodeWorld project before you dare to update or change it. If third-party code is in a submodule, you have to know how to interact with submodules, and you will usually also have to actually run its build system to get the file you really want.

    On the other hand, if third-party code is retrieved at build time from canonical sources, it's obvious, even to a new contributor, that you're using the canonical version. With the proposed directory structure, it's also self-documenting where each file originally came from, and self-evident that the file is mirrored code.

3. It's easy to teach machines to understand it as well (and in fact the structure is auto-generated by wget, which is already a dependency).

    I thought that some operators would rather *not* serve stuff like JQuery from their own machines. My personal hangups aside, it *is* more common to have pages take that stuff from canonical locations. This way, you can switch everything back to the canonical sources with a trivially obvious sed one-liner. It would also be easy to write code to do it as a build option, configuration option, or command line option. None of those are true if path names don't directly reflect  the origins of the files.

4. It's safer legally. By not distributing any of the code at all, you avoid the risk of Google's legal department missing something in one of the licenses (or some subtle interaction between that license and something else you do that they haven't noticed or thought about). That protects both the project and Google. You also avoid the project's getting entangled with legal disputes and positions that Google might want to take on for unrelated reasons. I don't think Google's lawyers probably see it as a big part of their job to care about the CodeWorld project.

5. ... and, OK, yes, it also means there's no need to guess at the internal rules of an uninvolved company that's opaque to me and to other potential contributors.

    From my point of view, that *eliminates* a lot of complexity. Repeatedly guessing until you chance onto something that an uninspectable script approves of doesn't strike me as a simple approach, for now or for the future. The rules aren't actually available outside Google. They  may change at any time based on unpredictable internal concerns. If Google's like most companies I've worked with, even the internally available documents are probably pretty bad anyway. 

    I'd also like to point out that this is *within* Google's guidelines as far as I know or can guess, in letter *and* spirit. Certain risks arise if you redistribute code. Google wants to protect you and themselves against those risks. If you don't redistribute code to begin with, then those risks never arise. I guess you could call that "working around the guidelines", but I'd call it "a simpler and surer solution to the problem they're trying to address".

> Doesn't this exclude your own scripts from source control? I'm confused.

It only prevents Git from suggesting checking in things that don't already exist under source control. Source control still applies to anything that *is* checked in, including the script. Things like "git commit -a" still work correctly.

... but I admit that it's confusing if you don't know that, and it also has the failure mode that Git won't know if you add a new file that *should* be checked in. I was kind of nervous about doing it that way. The problem goes away anyhow if the stuff is downloaded into the build directory.

> If you're going to do it this way, can we do it at build time? This is okay if it takes a long time, but rerunning install.sh isn't very common, and if it's quick., I'd rather just do it at build.
[... and ...]
> Anything fetched or built during install/build should go in the build directory.

I wasn't really sure which script to put it in or where best to put the files, and have no problem with those changes.

The download doesn't even happen after the first run unless you've cleaned your tree; it won't retrieve a file that it already has. Even if you do re-download everything, the whole process isn't much if any slower than a Web page load, since that's really what it's replicating.

The only possible thing to consider is that if people frequently tend to clean their build trees, and/or if install.sh uses the network and build.sh doesn't, then it gets harder to maintain an instance on an isolated network, because you end up having to hook it up to the Internet all the time to re-download the resources. But I don't know if either of those is true.

It's OK to symlink from "web" into "build", then?

cdsmith:
Okay, I'm persuaded.  In particular:

1. It's clearer where the scripts came from, and clearer how to add a new dependency, which makes future maintenance easier.
2. Should some dependency have a critical flaw of some kind, it's likely to be removed or updated in-place at the canonical source, and this will detect that, at least for new installs.  Vendoring the script risks distributing the faulty dependency forever.

I just want to be clear that we're choosing to do it this way for technical reasons, and *not* to avoid policy compliance.

cdsmith:
Yes, it seems like the right thing to do is to perform these downloads in build.sh, place them in `build/mirrored`, and symlink from `web/mirrored` to `build/mirrored`.

About your concerns, `build.sh` already performs network I/O.  At least, I know that it runs `cabal install`, which downloads from Hackage.  There's no reason not to add more network I/O during build.  It's also not common to clear the `build` tree.  Doing so leaves you waiting on install for an hour or two, since it would then need to rebuild and boot the patched GHCJS, among other things.

jbash:
I think the commit I just pushed addresses everything.

PR #1413, opened 2020-08-15
Delay dismissing 'compiling' message until the program actually starts running

### LINKED ISSUE #1390: Make is clearer when the program is loading
Right now it appears as if some compiled programs run very slowly, but much of that waiting time is spent downloading the compiled code.  To give the end-user a better feel for the behavior of their own code, it would be better if the environment made it clear when the code was done loading and actually started.

Fortunately, this isn't too difficult!  The compiled code already sends a `programStarted` message via `window.postMessage` when the code is fully loaded and starts to run.  Right now, that's used only for a timer that decides how to report deferred errors.  We could, instead, continue to show the compile wait dialog and dismiss it only when `programStarted` is received.  Would also have to be careful to always dismiss when something goes wrong.

Powell-v2:
Closes #1390

### DIFF
--- a/web/js/codeworld.js
+++ b/web/js/codeworld.js
@@ -955,6 +955,8 @@ window.addEventListener('message', (event) => {
         showRequiredChecksInDialog(msg);
       }, 500);
     }
+
+    sweetAlert.close();
     break;
   case 'showGraphics': {
     const runner = document.getElementById('runner');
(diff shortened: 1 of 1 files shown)


web/js/codeworld.js
```diff
@@ -1121,24 +1123,26 @@ function compile() {
     allowEscapeKey: false,
     allowEnterKey: false,
   }).then(() => {
-    window.cancelCompile();
+    stopRun();
   });
 
   sendHttp('POST', 'compile', data, (request) => {
     if (compileFinished) return;
-    sweetAlert.close();
```
cdsmith, line 1129: I think you will need to do this later if there are compile errors (see line 1181 where error messages are displayed when `success` is false`)

Powell-v2: Oh yeah, that's true, I missed that. Thanks!

PR #1414, opened 2020-08-15
Restore scroll position in guide

Powell-v2:
This functionality was inadvertently broken when nested accordions were introduced.

### DIFF
--- a/web/js/doc.js
+++ b/web/js/doc.js
@@ -350,14 +350,14 @@ window.onscroll = (event) => {
     const accordionOptions = {
       collapsible: true,
       heightStyle: 'content',
-      beforeActivate: (event, ui) => {
-        position.scrollLeft = 0;
-        position.scrollTop = 0;
-
+      beforeActivate: function loadArticle(event, ui) {
         const title = ui.newHeader.text();
         const path = title && getPath(shelf.named, title);
         if (path) {
           loadPath(path);
+
+          position.scrollLeft = 0;
+          position.scrollTop = 0;
         }
       },
     };
--- a/web/js/doc.js
+++ b/web/js/doc.js
@@ -99,12 +99,12 @@ window.onscroll = (event) => {
 
           if (linkable && clickable) {
             pre.classList.add('clickable');
-            pre.onclick = () => {
+            $(pre).on('click', () => {
               if (env) {
                 savePosition();
                 env.postMessage({ type: 'loadSample', code: text }, '*');
               }
-            };
+            });
           }
         })();
       }
@@ -115,13 +115,13 @@ window.onscroll = (event) => {
     const elems = root.getElementsByClassName('collapsible');
     for (let i = 0; i < elems.length; ++i) {
       const elem = elems[i];
-      elem.onclick = () => {
+      $(elem).on('click', () => {
         if (elem.classList.contains('expanded')) {
           elem.classList.remove('expanded');
         } else {
           elem.classList.add('expanded');
         }
-      };
+      });
     }
   }
 
@@ -136,7 +136,7 @@ window.onscroll = (event) => {
       iframe.setAttribute('frameborder', '0');
       iframe.setAttribute('scrolling', 'no');
 
-      iframe.addEventListener('load', (e) => {
+      $(iframe).on('load', (e) => {
         const currentTarget = e.currentTarget;
         const contentWindow = currentTarget.contentWindow;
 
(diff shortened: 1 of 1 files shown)


web/js/doc.js
```diff
@@ -99,12 +99,12 @@ window.onscroll = (event) => {
 
           if (linkable && clickable) {
             pre.classList.add('clickable');
-            pre.onclick = () => {
+            $(pre).on('click', () => {
```
cdsmith, line 102: Why is all this extra jQuery needed?

web/js/doc.js
```diff
@@ -350,14 +350,14 @@ window.onscroll = (event) => {
     const accordionOptions = {
       collapsible: true,
       heightStyle: 'content',
-      beforeActivate: (event, ui) => {
-        position.scrollLeft = 0;
-        position.scrollTop = 0;
-
+      beforeActivate: function loadArticle(event, ui) {
```
cdsmith, line 353: This seems unusual.  I'm surprised it's not a syntax error.  Any reason you're adding a name?

web/js/doc.js
Powell-v2: jQuery isn't strictly necessary. I think using it sparingly leads to a more readable code on certain occasions. At first, I wanted to replace `onclick` with a more flexible `addEventListener`, but then decided to opt in for jQuery's shortcut as it looks cleaner to me.

web/js/doc.js
Powell-v2: > I'm surprised it's not a syntax error

It was an arrow function before, and that typically means that an old-fashioned counterpart could be also used, if needed.

I only wanted to be more descriptive of what's happening inside this function. In general, I tend to assign meaningful names to functions as it helps me with debugging and makes code more readable.

web/js/doc.js
Powell-v2: Why would you want to remove it?

web/js/doc.js
Powell-v2: > to avoid polluting the scope

What do you mean by that, exactly?

web/js/doc.js
cdsmith: I always lean toward removing dependencies when it's possible to do so.  Dependencies load huge blocks of code which delay the page load, and much of it isn't needed.  They accumulate as cruft.  They take maintenance work to keep up-to-date for security fixes... and in practice, that doesn't happen.  (For example, it's come up recently that `blocks.html` in this project loads two different jQuery versions, and we're guessing at what that actually does.)  jQuery in particular does some very hard work for trivial things (e.g., `$elem.css(...)`, for example) that now have much faster native implementations in modern browsers.

Though I'm not familiar with the details, the general consensus among web developers I have interacted with is that modern DOM has made many of the features of jQuery obsolete, and it's just a legacy way to do things that are now built in.  There's `querySelectorAll` to find nodes.  There are lots of new DOM functions to manipulate those nodes.  There's `map` and `forEach` and arrow functions to easily operate on entire collections of nodes in even more flexible ways than jQuery offers.  There's `addEventListener` to add multiple listeners for the same event.

Finally, in practice, CodeWorld doesn't use one or the other consistently.  There are jQuery parts, and there are native DOM parts, and that actually makes it more difficult to contribute to than either one by itself, because now you need to be comfortable with **both**.  There's a lot of appeal in making this more consistent, and for the reasons above, if I went one direction or the other, I'd want to move away from jQuery.

Here's the counter-argument, though.  We're using at least two key components (the accordion, and jqtree) that are designed around jQuery.  So that already pulls in the jQuery dependency, and relies on jQuery-isms to interact with it.  So that's a bit of a sunk cause.  Unless those components were replaced, now isn't the time to try to remove jQuery from the code.  So, I don't care enough to block development.

Honestly, the main reason I mentioned it in this pull request was that it added a bunch of changes that didn't seem to add value, and made it harder to identify the important changes for review.

web/js/doc.js
cdsmith: I mean that named functions get dropped into the function-level scope, similar to `var` declarations.  That's another name the developer has to remember when they are reading the code, in case it gets called later.  When you write an arrow function, it's clear that no one is going to call it by its name.

web/js/doc.js
Powell-v2: I'm not particularly attached to jQuery and not a big fan of the imperative approach. Nevertheless, despite the fact that it might be a "legacy" way of doing things, jQuery provides a lot of utility. I was suprised to learn few months ago that it still dominates web development in 2020, surpassing any other framework (such as React and Angular).

I would venture a guess that some people who are comfortable with jQuery don't have a good grasp on the underlying mechanics, while devs who can manipulate DOM with the native API can quickly pick up jQuery, if need be.

I agree that dead code does no good for the application. At the moment, nothing could be done to alleviate this concern. But it the future, once all assets are preprocessed, we will have an opportunity to eliminate any unused code from the final bundle.

web/js/doc.js
Powell-v2: Got it, thanks for clarifying!

PR #1417, opened 2020-08-16
Remove inline js and css

### LINKED ISSUE #1372: Move inline JavaScript to its own files
In several places we have significant amounts of inline JavaScript in html files.  Aside from just being bad style, ESLint won't check them.  They should be moved to separate files.

Powell-v2:
Closes #1372

### DIFF
--- a/web/blockframe.html
+++ b/web/blockframe.html
@@ -46,48 +46,6 @@
     <div id="blocklyDiv" style="height: 100%; width: 100%;"></div>
 
     <script src="../js/blocks_runmain.js"></script>
-    <script>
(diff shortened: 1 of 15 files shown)


Powell-v2:
What is [this file](https://github.com/google/codeworld/blob/master/web/blocktest.html)? Doesn't seem to be used anywhere.

Powell-v2:
Not sure how that happened, but examples from the gallery aren't loading anymore. Could you advise? Here's the error I'm getting:

image

Powell-v2:
Sure, will do.

PR #1419, opened 2020-08-17
Fiverr

Gravity-Well:
First one 3rd of Unit 1


web/help/GuideUnit1.md
```diff
@@ -124,14 +121,18 @@ nametag = lettering("Camille")
...
 picture with letters (or any other kind of text) on it.
 
+Did you notice the word **Compiling** popped up and went away? This is the computer reading your instructions to make sure they are correct before executing them.
+
 Next, you can add a border to your nametag.  You might be tempted to add
 a new line like `nametag = ...` to your code, but you can't! Remember,
 your code is like a dictionary, and each definition in it should give
 the whole definition for that word.  To include a second shape in your
-nametag, you'll use **`&`**, which you can read as "and" or "in front
+nametag, you'll use an *operator*, **`&`**, which you can read as "and" or "in front
```
cdsmith, line 130: It might be a good idea to introduce operators here.  If so, though, I think a lot more needs to be said.  Students should be reminded of the operators they already know about on numbers, like addition.

This is already discussed in a collapsible box under the "Expressions and types" section later on.  I think it's a bit of a balance.  Right here, we're in a hurry to get students to writing their own creative programs as soon as possible, so whatever absolutely needs to be said should go here, but any long-winded asides can wait until later.

web/help/GuideUnit1.md
```diff
@@ -186,7 +187,7 @@ Try these examples to learn more:
     the computer will be confused and think you're defining a new
     variable.  This can cause a `Parse error` message.
 
-Once you've understood these examples, try your own combinations as well.
+Once you've understood these examples, try changing the numbers or making other changes.
```
cdsmith, line 190: Hmm.  I'm usually not a fan of the "change some numbers" suggestion, because it's giving students an easy out to just do random stuff without any specific intention, and check off the box.  I very much want students to add their own shapes, and hopefully even imagine and try to reproduce some design.

This might be a two-step process, though.  I think it would be an improvement if you ask them to just try some modifications as you've done here, but then follow up with an explicit exercise to do something more deliberate.  The exercise should have three steps: (1) Reflect on what you know how to do (answer: draw text, and different sized circles and rectangles, but only at the center of the screen., (2) Draw a name tag on paper using only these kinds of shapes, (3) create a program that displays a picture of what you drew.  If I were teaching, I'd have students submit their paper drawings (or photos thereof, given COVID stuff) as well as their programs, to verify that they made a goal and decided how to accomplish the goal, instead of just doing things at random.

web/help/GuideUnit1.md
```diff
@@ -355,9 +357,7 @@ favoriteColor = blue
 
     More on that later!
 
-Each of these lines is an **equation**, which says that two expressions are
-*equal*, or have the same value.  In math, you use equations in many
-ways, but in CodeWorld they are used specifically to *define variables*.
+Each of these is a type of **equation**, which says assign the value on the right to the name on the left. You have seen this type of equation in math, often with variables x and y. For example, in math, you might see x=10 and y=20.
```
cdsmith, line 360: As above, we need to be very careful to give a mathematically correct definition of an equation.  Then, as a second step, explain which specific kinds of equations CodeWorld can understand.

web/help/GuideUnit1.md
```diff
@@ -517,6 +517,16 @@ letters of the function you want, you can narrow down the list.  The type
 signatures tell you what types of information you need to provide to apply
 the function and what type you can expect to end up with.
 
+**Important**: The names of functions are *reserved*. That means you cannot use them as variable names. You would not be able to write program like this:
```
cdsmith, line 520: This would be a great time to use a warning callout box.  The syntax looks like this:

```
!!! Warning: Title goes here
    Text goes here.  Each line should all be indented 4 spaces.  You can put code blocks and everything.
```

I'm also not a huge fan of using "reserved" here.  I think there's an easier idea: a variable name can only refer to one thing.  Since the built-in functions are already defined by CodeWorld itself, students are not allowed to use the same name for their own variables.  This is exactly like how they cannot define the same variable twice in their own code.  There's no reason to make these different ideas, when they are really the same idea.

web/help/GuideUnit1.md
```diff
@@ -517,6 +517,16 @@ letters of the function you want, you can narrow down the list.  The type
...
+
+program=drawingOf(circle)
```
cdsmith, line 522: Surround your code with five tilde characters to put it in a box with the same coloring and fonts as the editor.  Like this:

```
    ~~~~~
    code here
    ~~~~~
```

web/help/GuideUnit1.md
```diff
@@ -796,7 +805,14 @@ the point (0, 0) -- on the coordinate plane, so you should measure your x
 and y distances from there.  As you define your own pictures, it's a good
 idea to continue this practice.
 
-For example, suppose you wanted a circle representing the sun in the top left
+To see a circle with radius 5 drawn on the corrdinate plane, use this code:
+
+    ~~~~~ . clickable
+    program = drawingOf(coordinatePlane & circle(5))
+    ~~~~~
```
cdsmith, line 812: I don't believe you want to indent these?

web/help/rlt-notes-unit1.md
```diff
@@ -0,0 +1,14 @@
+If you do not leave a space before and after the & the popup does not show
```
cdsmith, line 1: I'm not sure what you mean by "the popup".  In general, I don't think spaces around operators like `&` should make a difference, so I'd love to see what example you're thinking of here.

web/help/rlt-notes-unit1.md
```diff
@@ -0,0 +1,14 @@
...
+If the student makes change and clicks an example they get a warning. It might be beneficial to mention something about this and/or saving.
```
cdsmith, line 2: I do think it might help to address some of these UI issues when they come up, yes.  I'd keep it in callout boxes so it doesn't interrupt the flow of the main exposition.

web/help/rlt-notes-unit1.md
```diff
@@ -0,0 +1,14 @@
...
+
+Should variable really be constant since you cannot re-assign it a new value?
```
cdsmith, line 4: Variables are commonly used in mathematics, and can never be re-assigned new values.  We should call them variables.

web/help/rlt-notes-unit1.md
```diff
@@ -0,0 +1,14 @@
...
+Ex. program=xxx
+nameTag=xxx
+nameTag=yyy
+
+Might want to mention the red x error shows when using a variable before it's defined.
```
cdsmith, line 9: Sure, if there's somewhere that it would help to explain how the system indicates errors, go for it.  I think you should focus first on explaining the bigger idea, that errors will often be underlined in red as you type, and then explain the specifics of that idea from that basis.

web/help/ztest.md
```diff
@@ -0,0 +1 @@
+test file
```
cdsmith, line 1: Did you mean to delete this test file?

web/help/GuideUnit1.md
```diff
@@ -47,26 +47,23 @@ Here's the code you just wrote and what its parts mean.
     | My program | is  | a drawing of | the CodeWorld logo. |     |
 
 * `program` is the **variable** that you're defining. A variable is a name for
-  something.  In most math, variables are just one letter long, and they stand
-  for numbers.  In CodeWorld, though, variables can name many different types of
-  values: numbers, pictures, colors, text, and even whole programs.  Because
-  you will use so many of them, you can name variables with whole words,
-  always starting with a *lower-case* letter.
+  something.  Usually in math, variables are just one letter long and store
```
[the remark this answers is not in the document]
Gravity-Well: I understand your comment with regards to the goal of CodeWorld. I'm not sure, however, if you asking me to change something here or just understand the approach.

web/help/GuideUnit1.md
```diff
@@ -47,26 +47,23 @@ Here's the code you just wrote and what its parts mean.
...
+  numbers.  In CodeWorld, though, variables can store many different types of
+  things: numbers, pictures, colors, text, and even whole programs.  Because
+  you will use so many of them, you can name variables with whole words. There are some rules for naming variables. Most importantly, they must start with a *lower-case* letter and cannot contain spaces.
```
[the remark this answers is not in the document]
Gravity-Well: Definitely. I did just let it auto-wrap.

web/help/GuideUnit1.md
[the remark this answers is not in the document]
Gravity-Well: I understand your comment with regards to goal of CodeWorld. However, I'm not sure if you are asking me to change something or just understand this.

web/help/GuideUnit1.md
```diff
@@ -47,26 +47,23 @@ Here's the code you just wrote and what its parts mean.
...
 
 !!! collapsible: Camel case
     Sometimes, you may want more than one word to name a variable!  The
-    computer needs each variable to be a single word starting with a lower-case
-    letter--so leave out the spaces.  To make it easier to tell when a new word
-    starts, you can capitalize the *second* and *later* words.
+    computer needs each variable to be a single word or collection of words starting with a lower-case
```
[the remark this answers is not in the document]
Gravity-Well: Agreed. My goal was to emphasize the point since it is so important. But it your wording is better, especially for the age group.

web/help/GuideUnit1.md
```diff
@@ -47,26 +47,23 @@ Here's the code you just wrote and what its parts mean.
...
     ![](camel.png width="30%")
 
-    In your first programs, `drawingOf` and `codeWorldLogo` were written in
+    In your first program, `drawingOf` and `codeWorldLogo` were written in
     this way.  It's often called **camel case**.  Why?  Because the variable
     name has humps!
 
-* The **equal sign** means "is" and tells the computer that two expressions mean
-  the same thing.  It is used to connect a variable with its definition.
+* The **equal sign** means "is" and assigns the expressions on the right to the variable on the left. The variable must always be on the left.
```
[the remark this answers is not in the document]
Gravity-Well: Agreed. Again, I'm forgetting the goal of CodeWorld. I'm thinking to much in the programming realm and not the math realm. **Do I need to change it back or is that a change can be rejected from your end?**

web/help/rlt-notes-unit1.md
Gravity-Well: Ignore it. I thought I noticed that when there were no spaces around the &,, even though the program ran, the popup explanation that appears above the & when the mouse hovers over was not showing. I was mistaken. It shows.

web/help/rlt-notes-unit1.md
Gravity-Well: Agreed. Again, me thinking in programming terms like const.

web/help/ztest.md
Gravity-Well: I did mean to. It was just to make sure comited and pushed okay.

web/help/rlt-notes-unit1.md
Gravity-Well: I'll give it some thought.

web/help/GuideUnit1.md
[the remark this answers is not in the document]
cdsmith: I definitely don't think you should talk about variables storing things, as you do here.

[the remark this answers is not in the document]
cdsmith: I definitely don't think you should talk about variables storing things, as you do here.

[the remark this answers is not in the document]
Gravity-Well: Got it. reverted back to your wording, that it is just a name.

web/help/GuideUnit1.md
```diff
@@ -124,14 +121,18 @@ nametag = lettering("Camille")
 You've used a new function: **`lettering`**.  This function describes a
 picture with letters (or any other kind of text) on it.
 
+Did you notice the word **Compiling** popped up and went away? This is the computer reading your instructions to make sure they are correct before executing them.
```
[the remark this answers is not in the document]
Gravity-Well: Got it. Keep it math focused and not try to explain the programming part.

web/help/GuideUnit1.md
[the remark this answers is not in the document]
cdsmith: You should info the change from your side.  There only thing I can do is lie or reject the entire pull request, so I cannot merge it until corrections are made.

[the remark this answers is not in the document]
cdsmith: You should info the change from your side.  There only thing I can do is lie or reject the entire pull request, so I cannot merge it until corrections are made.

web/help/GuideUnit1.md
Gravity-Well: Noted

web/help/GuideUnit1.md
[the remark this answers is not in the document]
Gravity-Well: That's what I figured.

web/help/rlt-notes-unit1.md
```diff
@@ -0,0 +1,16 @@
+Did you mean to reference a butterfly but not draw one?
+I understand your point. Just wondering if the student might be
+expecting a butterfly? Probably not if they understand polyline
+at this point but just making sure.
```
cdsmith, line 4: I do see the possibility of confusion.  If you can explain the same idea and avoid that, go for it!

Gravity-Well: Added a very simple butterfly. It's on the remote repo, but won't show up when using the preview link you sent. I am I doing something wrong?

 https://code.world/doc.html?path=https://raw.githubusercontent.com/Gravity-Well/codeworld/fiverr/web/help/GuideUnit1.md

cdsmith: No, this is right.  Clickable code is the one thing that won't work in the preview URL.  That's because you are just looking at the guide, without the main CodeWorld window behind it.  Once I merge the pull request and students read it using the CodeWorld site, it will work.

If you really want them to see it inside the guide itself, then you can do that with an iframe.  You can search for `iframe` in the file, and copy one of those lines.  Replace the URL with the one you get by choosing "Share Without Code" from the Share dialog in CodeWorld.  That will actually run the code inside the Guide itself.

cdsmith: As for this particular usage, I think this loses the point of the example.  The idea was to start with just the coordinate plane, then add points one by one.  If you just show them the completed program, I don't think it communicates that very well.

Gravity-Well: Maybe if I reduced it to part of the butterfly. For example, part of the left wing. I'm just trying to make a bit more than an empty polyline. 
After showing part of the left wing, explain to them how they could remove the coordinate plane, view it, put it back in and add more. Thoughts

cdsmith: I suspect the right thing to do here is to include multiple steps, and show how they can add to the program one vertex at a time.  It should probably be in a callout box if it gets that long, just to delineate it well from the rest of the text.

Gravity-Well: I want to contribute in a meaningful way. I'd enjoy creating those steps if you think it would enhance the point. I think it would. But I can skip it and we can move on if you do not. Let's consider a video chat soon to discuss some of your ideas. The only real restrictions on my time are video calls I have on T,TH and F from 10:30-2:00 EST. Other than that, tasks can be shuffled.

cdsmith: Yes, as I mentioned, I do think more detail here would be an improvement.  Just put it in a callout box so it doesn't interrupt the flow too much.  Will contact you elsewhere about setting up a video call.

web/help/GuideUnit1.md
```diff
@@ -1896,12 +1894,37 @@ y axis, since those are the angles in between the numbers -90 and 90.  But the
 expression `arc(270, 90, 5)` covers the angles to the left of the y axis,
 because those are the numbers between 270 and 90.
 
+To see a half circle with radius 5 drawn on the corrdinate plane, use this code:
+
+~~~~~ . clickable
+program   = drawingOf(halfCircle & coordinatePlane)
+halfCircle=arc(-90,90,5)
```
cdsmith, line 1901: Minor nit: I'd like to encourage students to use spaces around the equal sign, so let's at least model that for them.

web/help/GuideUnit1.md
```diff
@@ -1896,12 +1894,37 @@ y axis, since those are the angles in between the numbers -90 and 90.  But the
...
+thickness of the line to draw.
+
+To draw half of a tire, use this code:
+
+~~~~~ . clickable
+program   = drawingOf(halfTire & coordinatePlane)
+halfTire=thickArc(-90,90,5,4)
+~~~~~
+
+Notice the thickness is split between both sides of the arc.
```
cdsmith, line 1915: I'm not sure it will be clear what you mean by this, and you may need to be more explicit and refer to the coordinate plane: for example, by pointing out that when the curve intersects the x axis, it is centered around an x value of 5, but it extends all the way from 3 through 7.  It could be useful to copy the image with the coordinate plane, and show it here.

(Copying an image is a little tricky.  Basically, you want to run the program, right-click on it, then wait for the on-screen controls to go away before choosing "Save image as..." in your browser.  Then you can add the file to your git branch in the `web/help` directory, and then refer to it from Markdown.  Alternatively, you can use the iframe pattern to just run the program inline.  Either will work, though I normally try to save the image unless I am going for some kind of animation or interactive demo.)

web/help/rlt-notes-unit1.md
Gravity-Well: I do plan to do the incremental butterfly. Just did not include it in this push.

web/help/GuideUnit1.md
Gravity-Well: Gotcha

web/help/GuideUnit1.md
Gravity-Well: Yeh. I was having trouble with the wording but though it was important to point out how the thickness worked. I'll add more of a description and an image.

PR #1434, opened 2020-09-06
Always show share button

### LINKED ISSUE #1410: Show the Share button even when the code isn't running
Suggested by @jbash in #1409.  There should still be a Share button to share code even when the code isn't running.

In terms of implementation, this means that the code will have to be sent to the server when Share is clicked, since it hasn't been sent yet.  Currently the only way to do that is to invoke `/compile` on the server.  That is wasteful, though, so I should probably add a new server-side endpoint to store code without trying to compile it.

Open questions:

1. What if the current code has been modified since the program was run?  Currently, we show a warning, and offer to re-run the program (which will then cause Share to share the latest code as well).  Instead, if users shouldn't think of sharing as connected to running, maybe we should just ask them which they intended to share.
2. If this might cause more shares of incomplete code, should we stop users from accidentally sharing code that hasn't been run without source access?  My sense is probably no... but it's worth considering if this would be confusing.  The risk is that students will share something that they think works without testing, and then the person they are sharing with only sees a white box, and the student is frustrated because they never saw the error message.

Powell-v2:
Closes #1410


cdsmith:
@googlebot I fixed it.

PR #1519, opened 2021-09-13
Stylish Haskell config

peterbecich:
This Stylish Haskell configuration formats the code similarly to how it already is. I've formatted two files to show the differences. There may be other differences, as well.

The differences could be reduced further by disabling the `simple_align` settings.

Footnote for Emacs users: I've tested this to work with `haskell-mode-format-imports` (https://github.com/haskell/haskell-mode/blob/886795c15036d566aeced66f9508ae61ec0287ec/haskell-mode.el#L1069)

### DIFF
--- a/.stylish-haskell.yaml
+++ b/.stylish-haskell.yaml
@@ -1,0 +1,56 @@
+steps:
+  - simple_align:
+      cases: always
+      top_level_patterns: always
+      records: always
+      multi_way_if: always
+
+  # Import cleanup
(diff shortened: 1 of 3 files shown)


cdsmith:
format.sh already formats with ormolu.  I'm not sure we want the two tools warring with each other.

peterbecich:
Oh cool, I didn't realize this project used Ormolu, thanks

PR #1520, opened 2021-09-13
[WIP] GHC 8.10 and Nix support

peterbecich:
Using https://github.com/input-output-hk/haskell.nix ; a.f.a.i.k. this is the easiest way to build a `cabal.project` with Nix

GHCJS is provided


cdsmith:
If you're going to update to GHC 8.10, that sounds awesome!  But I think it should be done in a minimal pull request of its own rather than combining with a bunch of CI changes.

peterbecich:
Sounds good
GHC 8.10.7 https://github.com/google/codeworld/pull/1522
Nix https://github.com/google/codeworld/pull/1523
both are works-in-progress

PR #1522, opened 2021-09-14
[WIP] GHC 8.10.7

peterbecich:
There is an issue compiling `codeworld-base`:

```
$ cabal build codeworld-base
...
dist-newstyle/build/x86_64-linux/ghc-8.10.7/codeworld-base-0.2.0.0/build/Extras/Cw.dyn_o )
ghc: panic! (the 'impossible' happened)
  (GHC version 8.10.7:
        expectJust failed to detect OverLit
  error, called at compiler/utils/Maybes.hs:57:27 in ghc:Maybes
  expectJust, called at compiler/GHC/HsToCore/PmCheck.hs:551:16 in ghc:GHC.HsToCore.PmCheck

Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug

```

Cabal 3.6.1.0
GHC 8.10.7
Debian Bullseye


cdsmith:
Hmm, that's too bad.  The good news, though, is that CodeWorld already builds its own patched version of GHCJS, if someone is willing to backport this fix for us.

cabal.project
```diff
@@ -10,3 +10,6 @@ packages:
   codeworld-server/
   codeworld-base/
   funblocks-client/
+
+constraints:
+  ghcjs-dom ==0.9.*
```
cdsmith, line 15: I'd rather keep all the version constraints in one place.  Can this be kept in the cabal file?

codeworld-api/codeworld-api.cabal
```diff
@@ -61,7 +61,7 @@ Library
                        mtl                  >= 2.2.1 && < 2.3,
                        random               >= 1.1   && < 1.2,
                        ref-tf               >= 0.4   && < 0.5,
-                       reflex               >= 0.6.3 && < 0.7,
+                       reflex               >= 0.6.3,
```
cdsmith, line 64: I believe we need upper bounds on versions to upload to hackage.  So feel free to relax bounds, but let's leave them in place for codeworld-api.  Other packages don't need upper bounds, since they don't get released to hackage.

codeworld-compiler/codeworld-compiler.cabal
```diff
@@ -54,16 +54,18 @@ Library
     directory,
     exceptions,
     filepath,
-    ghc == 8.6.5,
-    ghc-boot-th == 8.6.5,
+    ghc == 8.10.*,
+    ghc-boot,
+    ghc-boot-th == 8.10.*,
+    ghc-lib-parser,
```
cdsmith, line 60: I don't see where this is used.

If it's needed, ghc-lib-parser changes dramatically between versions (it's just a copy of parts of ghc), so it needs version bounds.  But it would be better to leave it out and update the ghc code, if it's not too difficult.

codeworld-compiler/src/CodeWorld/Compile/Framework.hs
```diff
@@ -305,32 +311,95 @@ ghcParseCode extraExts src = do
...
   return $ case GHC.unP GHC.parseModule state of
     GHC.POk _ (GHC.L _ mod) -> GHCParsed mod
-    GHC.PFailed _ _ _ -> GHCNoParse
+    GHC.PFailed _ -> GHCNoParse
 
 fakeDynFlags :: GHC.DynFlags
 fakeDynFlags = GHC.defaultDynFlags fakeSettings fakeLlvmConfig
 
+fakePlatformMisc :: GHC.PlatformMisc
+fakePlatformMisc = GHC.PlatformMisc
```
cdsmith, line 320: I almost cringe to ask it, but can you include field names here so the reader has some hope of finding what they need to change in the future?

codeworld-server/codeworld-server.cabal
```diff
@@ -39,7 +39,7 @@ Executable codeworld-server
     fast-logger,
     filelock,
     filepath,
-    haskell-src-exts < 1.21,
+    haskell-src-exts,
```
cdsmith, line 42: Thanks!  I looked up the history, and this was originally added to keep hindent working, but CodeWorld now uses Ormolu instead of hindent, so dropping the restriction is fine.

peterbecich:
Here is the backport; it solves the error: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/6554

cdsmith:
Cool!  I'm not sure whether there will be a new GHC release in the 8.x series, but if you make a patch (test case not necessary) and drop it in https://github.com/google/codeworld/tree/master/ghc-artifacts, and then apply the patch in install.sh near https://github.com/google/codeworld/blob/master/install.sh#L226, then CodeWorld will build with the change.  That should unblock you.

Speaking of patches there, I wonder whether the remaining patches in that directory are still needed.  The default main patch was contributed upstream, and has probably made it into GHC 8.10, but I'll verify that.  The dedup fix for the linker was a GHCJS patch, and my hope is that it's also merged in the 8.10 branch.  Again, I'll try to find evidence for this and report back.

cdsmith:
I've confirmed that the default main patch does NOT need to be ported.  It's already part of GHC 8.10.  Unfortunately, https://github.com/ghcjs/ghcjs/pull/788 still isn't merged, so this one needs to be ported forward.  There have been changes to that file, so it's not immediately obvious to me how to do so, but @3noch or @tomsmalley may know.

PR #1559, opened 2021-12-31
Remove local_only middleware in blank-canvas driver

3noch:
I see no reason to *prevent* one from using a non-local connection, particularly when the `local` middleware might not be entirely accurate. I'm using this change to allow my [Gitpod](https://github.com/3noch/haskell-gitpod-template/tree/codeworld) to show the drawing after saving files.


cdsmith:
Thanks!

PR #1566, opened 2022-01-15
fix non-GHCJS compilation

peterbecich:
With GHC 8.6.5 this succeeds locally for me:
```
% cabal build all
Up to date
```

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -10,3 +10,6 @@ packages:
   codeworld-server/
   codeworld-base/
   funblocks-client/
+
+constraints:
+  ghcjs-dom ==0.9.5.0
--- a/codeworld-auth/codeworld-auth.cabal
+++ b/codeworld-auth/codeworld-auth.cabal
@@ -31,7 +31,7 @@ library
                                           , directory
                                           , filepath
                                           , http-conduit
-                                          , jwt >= 0.10.0
+                                          , jwt == 0.10.*
                                           , snap-core
                                           , split
                                           , text


cabal.project
```diff
@@ -10,3 +10,6 @@ packages:
   codeworld-server/
   codeworld-base/
   funblocks-client/
+
+constraints:
+  ghcjs-dom ==0.9.5.0
```
cdsmith, line 15: This should really go in the cabal file (presumably in `codeworld-api/codeworld-api.cabal`) rather than here.  To be honest, I'm a little surprised that there's a cabal.project file checked in.  That seems wrong...

peterbecich: Regarding the `cabal.project` file, I know the primary way to build CodeWorld is with the scripts provided; needed for GHC patches, etc. IMO the `cabal.project` file is useful nonetheless. For one, I'm trying to get the whole project building with Nix https://github.com/google/codeworld/pull/1523 and `cabal.project` is needed for that.

cabal.project
```diff
@@ -10,3 +10,6 @@ packages:
   codeworld-server/
   codeworld-base/
   funblocks-client/
+
+constraints:
+  haskell-src-meta ==0.8.7
```
peterbecich, line 15: I don't think this can go anywhere else, because it is a transient dependency.

`haskell-src-meta` is only used directly by `codeworld-available-pkgs`, which is not in this `cabal.project` file.

peterbecich: I think `reflex` fails to build with `haskell-src-meta-0.8.8`, but only as a dependency of CodeWorld. I can't reproduce the error in `reflex` by itself.

The transitive dependency is here
https://github.com/peterbecich/reflex/blob/823afd9424234cbe0134051f09a6710e54509cec/reflex.cabal#L183-L189
```
cabal build all --flags="use-template-haskell"
```
succeeds with `haskell-src-meta-0.8.8`.

cdsmith: Okay, I'll try to reproduce the failure and understand what's going on.  If this is the case, though, I still think the constraint should be added in `codeworld-api.cabal`  By adding it in `cabal.project`, you make builds initially appear to succeed, but then the `codeworld-api` that gets deployed to hackage may still fail to build for users who get it from hackage.  Honestly, I'd rather know that it's broken than make it appear to work in development and then break when deployed.

peterbecich: Here is the error with `haskell-src-meta-0.8.8`:
```

src/Language/Haskell/Meta/Syntax/Translate.hs:493:19: error:
    Not in scope: data constructor ‘Exts.DerivVia’
    Perhaps you meant ‘Exts.Deriving’ (imported from Language.Haskell.Exts.Syntax)
    Module ‘Language.Haskell.Exts.Syntax’ does not export ‘DerivVia’.
    |
493 | toDerivStrategy d@Exts.DerivVia{}      = noTHyet "toDerivStrategy" "2.14" d
    |                   ^^^^^^^^^^^^^
cabal: Failed to build haskell-src-meta-0.8.8 (which is required by
test:unit-tests from codeworld-api-0.7.0, exe:codeworld-game-bot from
codeworld-game-server-0.1 and others). See the build log above for details.
```

peterbecich: I think this will solve it: https://github.com/haskell-party/haskell-src-meta/pull/32

